PR #857, opened 2015-09-08
Get Squeak tests passing 

dalehenrich:
turns out that the `Zinc-Seaside`package was included in #common with no `Zinc Project` defined for Squeak, since Zinc isn't ported to Squeak ... 

[Squeak tests are finally green](https://travis-ci.org/dalehenrich/Seaside/builds/79217133)


jbrichau:
Thx @dalehenrich for spotting that!

PR #866, opened 2016-06-04
Fix typo

Wilfred:


marschall:
Fixed in SmalltalkHub, should be fixed on GitHub with the next synch.

PR #867, opened 2016-06-05
change the version # to 3.2.0

pdebruic:
The version of #seasideVersion on Smalltalk hub is correct, but not here in github.  Not sure if this is the only difference between the two repos or not.


jbrichau:
Sorry for not noticing your pull request earlier... I now fixed this in #896

PR #905, opened 2017-01-14
Streaming patch

JupiterJones:
When attempting to use the Streaming server in pharo (and GemStone) there were a couple of minor issues with:

GRCountingStream>>javascriptOn: aStream (was missing)
WAResponse>>writeStatusOn: aStream (was adding characters which caused an error with WAComboResponse)

Also added is a new class in Seaside-GemStone-Adaptors-Zinc to support streaming in GemStone.

Annoyingly there are a bunch of other changes grouped in here by exporting from GemStone which removes extra spaces and end of line characters.


repository/Seaside-Core.package/WAResponse.class/instance/writeStatusOn..st
```diff
@@ -1,11 +1,12 @@
...
\ No newline at end of file
+writeStatusOn: aStream
+  aStream nextPutAll: self httpVersion.
+  aStream
+    nextPutAll: $  greaseString;
+    nextPutAll: self status greaseString.
+  self message isNil
+    ifFalse: [ 
+      aStream
+        nextPutAll: $  greaseString;
```
theseion: Is this really necessary? Not that it's a huge deal, but the iteration overhead for single character + an additional method activation seem a bit ugly.

repository/Javascript-Core.package/GRCountingStream.extension/instance/javascript..st
```diff
@@ -0,0 +1,3 @@
+*javascript-core
+javascript: anObject
+  anObject javascriptOn: self
```
theseion: This is necessary as GRCountingStream does not extend Stream (which would have the method).

repository/Seaside-Core.package/WAResponse.class/instance/writeStatusOn..st
marschall: I would prefer to leave it at #nextPut: as well

marschall:
@JupiterJones the PR seems to reformat a lot of methods, can you undo this as this makes in unnecessarily hard to review?

marschall:
- `GRCountingStream >> #javascriptOn:` was fixed by #938
- `WAResponse >> #writeStatusOn:`, "caused an error with WAComboResponse" can you be more specific? If `#nextPutAll:` with a collection of characters works then `#nextPut:` with a single character should work as well.

theseion:
I'll attempt to extract the actual and valid changes and update the PR.

theseion:
This PR has been superseded by #1075.
- The change to `#baselinefileSystem:` was already fixed in 8e8299409220c8f645da57f79beff9d97d1298c1
- The missing extension for `#javascript:` was already added in b9ce986589d7e088b945189d49f753da4bc8a2a0
- I've ignored the change to `#writeStatusOn:` as we don't have a clear error report

PR #914, opened 2017-05-19
replaced WADynamicVariable with GRDynamicVariable

theseion:
* sublcasses of WADynamicVariable now inherit from GRDynamicVariable
* removed WADynamicVariable
* removed WADynamicVariable tests


marschall:
I don't think `WADynamicVariable` should have been removed, it unnecessarily breaks users with no good migration path like Seaside-Gettext. I would leave it in as a subclass of `GRDynamicVariable `.

PR #919, opened 2017-05-27
Deleting session cookie with redirect can lead to infinite redirects

theseion:
* removed WAResponseGenerator>>expiredRegistryKey (relevant logic moved to WARegistry>>handleExpired:)
* added extensive comment to WARegistry>>handleExpired: on why the redirect upon expiration was replaced with delegation to the default behaviour


marschall:
I think this would need an entry in the changelog as it breaks the behaviour of existing applications. [Seaside330Changelog](https://github.com/SeasideSt/Seaside/wiki/Seaside330Changelog)

marschall:
I redid the PR because of merge conflicts, I changed

- I left #expiredRegistryKey in for people that want the old behavior
- I changed WARequestContext #newCookie to do the encoding there as well
- Updates the UTF-8 cookie tests as well
- Fixed tests by adding a subclass that introduces the old behavior

Issue #920, opened 2017-05-28
Make WAHtmlRoot configurable


theseion:
I need WAHtmlRoot to be configurable so I can override certain methods.

marschall:
Do you mean the one created in `WARenderPhaseContinuation`?

theseion:
Yes.

marschall:
You could always just provide your own render phase continuation, AFAIK this is already customizable.

theseion:
Of course! Stupid. Thanks for clearing that up.

PR #921, opened 2017-05-28
issue #920

theseion:
* replaced referene to WAHtmlRoot with dispatch to configuration
* added description for available html root classes to WARenderLoopConfiguration


theseion:
Invalid (see issue #920).

PR #922, opened 2017-05-28
WARequestCookie>>path: sets empty path for '/' 

theseion:
* initialize the cookie path attribute to '/' as that is what browsers are supposed to use as default / fallback
* the cookie path attribute will now always be set
* WARequestCookie>>pathEncoded now correctly answers '/'  instead of the empty string when '/' was set as path 
* cookie paths are now forced to use a leading slash. Browsers will replace paths without leading slash with '/'
* cookie paths are now forced to omit a trailing slash. Browsers will remove the trailing slash (with '/' as the exception)
* nil, the empty string, and '/' are now treated as equal when setting the path of a cookie

* added tests for all of the above
* updated other cookie tests that did not expect the path attribute in the output


marschall:
I redid the PR because of merge conflicts, I changed

- I left #expiredRegistryKey in for people that want the old behavior
- I changed WARequestContext #newCookie to do the encoding there as well
- Updates the UTF-8 cookie tests as well
- Fixed tests by adding a subclass that introduces the old behavior

PR #925, opened 2017-05-28
Added mapping to Prototype's `PeriodicalExecutor`

theseion:
* added PTPeriodicalExecutor decoration
* added PTObject>>every:

Note: although there are two different classes mapped to PeriodicalUpdater I have not touched them as the are copies of the old SUPeriodicalEvaluator and SUPeriodical; if there's a bug here it's now probably a feature...


marschall:
The Javascript class is named PeriodicalExecut**e**r but the Smalltalk class is named PeriodicalExecut**o**r.

What is the reasoning for adding #every: to PTObject? There seems to be not equivalent in the Javascript library
http://api.prototypejs.org/language/Object/

theseion:
Ported 1:1 from Seaside 2.8

marschall:
I would prefer the name PTPeriodicalExecuter and leaving out #every: in order to have a closer match with the JS library.

theseion:
I can live with the name change.  

I can also live with removing #every: but I'd prefer to keep it as we use it in existing applications.

jbrichau:
I would prefer the `PTObject>#every:` method to be put inside `PTFactory`. It seems it's a factory method for the Seaside binding.

jbrichau:
(disregard my earlier comment, I also deleted it already)
`PTObject>>every:` should actually even be implemented on `JSObject` as it can be sent to any `JSObject` instance representing JS code. There are numerous convenience methods there to wrap a `JSObject` instance with additional decorations, so it fits perfectly.

My only concern is that this is legacy and if people start using `every:` without realizing they need to have the prototype.js library, they are going to be confused. Then again... if you load the prototype extension... you should know what it is.

marschall:
@jbrichau convinced me

jbrichau:
@theseion what do you think? Can you add the necessary changes to this PR?

theseion:
Done. Sorry for the delay.

marschall:
I ended up redoing the PR because of merge conflicts. I kept it mostly the same except:
- I added some comments
- I added a test
- I removed the duration as it's serialized as milliseconds but seconds is expected

PR #926, opened 2017-05-28
WADynamicVariable should inherit from GRDynamicVariable

theseion:
* replace WADynamicVariable with GRDynamicVariable
* keep WADynamicVariable for backwards compatibility


jbrichau:
I finally managed to read this branch into an image with Iceberg :)

I do have a quibble with the commit: why is it that all the author metadata in (e.g. in `methodProperties.json`) has been changed even if those methods were not touched? I would prefer to have a clean commit only containing the changes necessary. If no metadata is necessary when using Iceberg, I would expect these not to have been changed? Keep in mind: I'm used to using filetree :)

Also, before merging this in, we need to fix the loading in older platforms and also the Gemstone part.

Though I don't know if it would be easier if we recreate this PR. I will experiment locally with Iceberg now to see what it does...

jbrichau:
Replaced with https://github.com/SeasideSt/Seaside/pull/936

PR #930, opened 2017-07-08
Fixes to make Seaside 3.2 load in GemStone 3.4

marianopeck:
Basically, I need the fixed in `baselinefilesystem:` change #squeakCommon to #common and `baselinegemstone:` to add platform attributes to 3.4.


marianopeck:
Looks like I made this PR wrong as I started from the 3.2 branch but then committed to master. I will re-apply my changes over master and re-commit.

Issue #978, opened 2017-12-02
Seaside does not load in Pharo 7


jecisc:
The baseline does not manage Pharo 7.

I'll probably do a PR to fix this.

PR #979, opened 2017-12-02
Adapt baseline to Pharo 7.

jecisc:


jecisc:
Close issue #978

jbrichau:
I merged the PR#980 for dev branch, but it shows we hit another issue in https://github.com/Metacello/metacello/issues/476 
Holding off merging this for master when that's fixed

svenvc:
I just tried to load Seaside 3.2 (GitHub.1516790362) in Pharo 7 (Pharo-7.0+alpha.build.493.sha.d53a70bc946ed83b046f61ac35f00c67fa77269a (32 Bit)) with the changes proposed here and at least the load and some quick testing with ZnZincSeasideAdaptor seems to indicate no direct problems.

Why don't we integrate this so that others can load Seaside 3.2 in Pharo 7 too ?

This is really important to make progress, IMHO.

jbrichau:
I can try again but we did hit issues: https://travis-ci.org/SeasideSt/Seaside/jobs/313946580
So, the load now works?

svenvc:
With the modifications described here (just add a couple of #'pharo7.x' tags here and there) the load worked. I did see the same error first, but not with the modified baseline loaded. I did do this load interactively, not as a CI build, using:
````
Metacello new
 baseline:'Seaside3';
 repository: 'github://SeasideSt/Seaside:master/repository';
 load.
````

jbrichau:
I'll take another look but Seaside builds on SmalltalkCI are failing for Pharo 7, so as long as this is the case, we cannot reliably move to Pharo7. Now, I understand we should at least make it load and I will merge now.

jbrichau:
Copied the changes of this PR in 3c7efa6500cf945efdad9455544979bdc3288736

jecisc:
@jbrichau There was more changes than this.

jecisc:
There was also: baselinejavascript..st, baselinerest..st and baselineslime..st

jbrichau:
@jecisc Yeah, noticed it too. Thanks for watching. I added another commit.
Mind that the `load: 'CI'` does not work because of a DNU on `Pragma`, so this needs to change as well if we want a build.

PR #985, opened 2018-01-28
GemStone 3.4.1 does not understand #_sharedCounter:setValue:

pdebruic:
I'm not sure if 3.3.X and earlier versions know #sharedCounter:setValue: so I don't want to mess with the BaselineOf yet but wanted this issue/fix recorded/shared.

### DIFF
--- a/repository/Seaside-GemStone-ServiceTask.package/WAGemStoneServiceTask.class/class/reset.st
+++ b/repository/Seaside-GemStone-ServiceTask.package/WAGemStoneServiceTask.class/class/reset.st
@@ -2,5 +2,5 @@ accessing
 reset
 
 	InProcess := Queue := nil.
-	System _sharedCounter: self sharedCounterIndex setValue: 1.
+	System sharedCounter: self sharedCounterIndex setValue: 1.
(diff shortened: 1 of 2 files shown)


dalehenrich:
Looks like #sharedCounter:setValue: was introduced in 3.2.x (present in 3.2.0 and not present in 3.1.0.6)

dalehenrich:
... and 3.1.0.6 does have #_sharedCounter:setValue: ... so that's most likely the time frame when that code was first introduced ...

pdebruic:
Hi  - So I don't know what to do here.  I found a problem in the BaselineOfSeaside3 where the Filesystem group is not defined for GemStone.  I've added the fix to my repo and this PR is updated, but I think I need to change what I'm diffing against because Dale also changed the BaselineOfSeaside3 since this PR was started so the metadata isn't correct in this PR.  

Is there a guide or description of what to do in this situation somewhere I could read?

pdebruic:
I think I can delete my fork of this repo, refork, and push the changes and have 2 prs.  So I'll do that.

PR #991, opened 2018-08-31
On-Input Event Handler

seandenigris:
Usage: `aCanvas jQuery this onInput: (aJavascriptScript asFunction: #(event))`
Motivation: Creates JS like the following: `$("#an-id").on('input', function(newVal) {  })`
Tests: Added to `#testEventsDom` and `#testRegister`, even though both tests seem to be amost exact duplicates.


marschall:
I redid the PR as #1010

- added metadata
- fixed the comment
- put the tests in the alphabetically correct order
- all the other events on jQuery instance are functions but `input` does not seem to be a function

PR #992, opened 2018-09-05
Update jQuery-UI from v1.10.4 to 1.12.1

seandenigris:
Re-upload JS file to File Libraries. All jQuery and jQuery UI tests passing


marschall:
I redid it with metadata as part of #1013

Issue #1053, opened 2018-11-26
Simplify gettext baseline


jecisc:
We can simplify the baseline of gettext since the minimal pharo version for this project is Pharo 4.

PR #1054, opened 2018-11-26
1053-Simplify-gettext-baseline

jecisc:
Since Pharo is the only one using gettext, merge Seaside-Gettext-Core and Seaside-Pharo-Gettext-Core
+ Simplify baseline


jecisc:
Fixes #1053

jbrichau:
I should check, because there is a gettext for Gemstone (we're using that one in Yesplan).
Dunno why it got slipped from the baseline...

jecisc:
Maybe it is in an external project as it is now in Pharo?

jecisc:
Oh I see that here we load the external project and have a compatibility package for Seaside for it. So it's probably not that.

jecisc:
Any news on this?

jbrichau:
Gettext is on my radar (because it's broken right now); but I'm trying to devote my spare time for Seaside to the website now.

jbrichau:
@jecisc Changed the base to master, but I think we can just close this one when I (or someone) address gettext

jbrichau:
Closing this PR as I'm fixing it on the gettext branch (which is to be merged when gettext upstream change has merged)

PR #1071, opened 2018-12-08
1070-Seaside-Tests-REST-is-never-referenced

### LINKED ISSUE #1070: Seaside-Tests-REST is never referenced

jecisc:
Remove Seaside-Tests-REST
Fixes #1070


theseion:
Same here, doesn't this package have a purpose, eve if it is not being referenced?

jecisc:
I guess that, as for swagger, we now have a package with “-Core” in the name. (I’m on Mobil so it’s hard to verify)

PR #1073, opened 2018-12-08
1072-Seaside-Swagger-and-Seaside-Tests-Swagger-are-never-referenced

### LINKED ISSUE #1072: Seaside-Swagger and Seaside-Tests-Swagger are never referenced

jecisc:
Remove Seaside-Swagger and Seaside-Tests-Swagger

Fixes #1072


theseion:
Out of curiosity, why did you remove these packages? Don't they have any purpose?

jecisc:
I think they were renamed, but they were migrated by the migration script. So currently it’s just dead code because the real packages have “-Core” in the name.

theseion:
Fair enough. I'll merge those PR's then.

PR #1079, opened 2018-12-20
Fix classes that are shown as "Uncategorized" in latest Pharo 7 by pu…

astares:
Fix classes that are shown as "Uncategorized" in latest Pharo 7 by putting them into a tag


theseion:
Thanks!

PR #1080, opened 2018-12-20
Add an Open button to open the webbrowser from control panel

astares:
Add an "Open" button to Seaside adapter control panel to quicky open a WebBrowser on the adapter itself


astares:
screenshot from 2018-12-20 01-30-10

astares:
Pharo 4 will not have git anyway, no?

jecisc:
Seaside is still loadable in pharo 4.

You don’t have Iceberg but FileTree still work.

astares:
Is this really a use case? Johan will convert to Tonel so we have Windows back on track once an issue with Tonel converter is sorted out.

Do we really have known users of P4 with FileTree?

jecisc:
Filetree will stay for a while because Squeak does not have a tonel support anyway and the repo is also for squeak.

astares:
@jecisc   It's a change in "Seaside-Pharo-Tools-Spec", not a Squeak one if I'm not mistaken.

jecisc:
Yes but we can't have one package in tonel format and the others in filetree.

astares:
Agree - but we will have a branch with Tonel

jecisc:
I think that the goal is not to split the development into two branches and to have differences between the two of them.

astares:
I guess the idea is to keep the development in file tree for now (until the rest catches up) - but nonetheless provide tonel converted versions (at least for the releases) so it could be loaded on Windows.

jecisc:
Yes, so the code should be the same between both, else it will make it really hard to sync.

jbrichau:
@astares I moved this PR to merge into master when accepted. I did not take a look at the code but I agree with @jecisc comments that the code should do feature detection so that it only gets added when the `WebBrowser` class is present.

theseion:
@astares Can you make the requested change?

theseion:
The new Spec2 based control panel includes such an open button. Closing this PR in favor of #1155 .

astares:
👍

PR #1083, opened 2018-12-22
1082-Merge-Comet-Core-and-Comet-Pharo-Core

### LINKED ISSUE #1082: Merge Comet-Core and Comet-Pharo-Core
From what I can read in the baseline, Comet is only loadable in Squeak and Pharo. Also, both Squeak and Pharo also load Comet-Pharo-Core.

We can then merge Comet-Core and Comet-Pharo-Core since those two packages are always loaded together.

jecisc:
Merge Comet-Core and Comet-Pharo-Core.

Fixes #1082


jecisc:
The Pharo 7 failure is not related. One of the latest Pharo change broke SmalltalkCI.

jecisc:
Apparently, the structure separating Pharo is to ease the port to other Smalltalk.

I'll close the PR.

PR #1096, opened 2018-12-23
Improve Seaside Welcome experience

jbrichau:
Now starting a default Zinc adaptor when loading the Welcome page _and_ no adaptor is present yet.

### DIFF
--- a/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselineadaptors..st
+++ b/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselineadaptors..st
@@ -27,10 +27,12 @@ baselineadaptors: spec
 						loads: #('Core');
 						repository: 'http://ss3.gemtalksystems.com/ss/WebClient' ].
 			spec
-				package: 'WebClient-Seaside-Adatpor'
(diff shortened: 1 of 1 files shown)


jbrichau:
I thought that if a port was already taken, it just does not work, but if this crashed on Windows, we might want to hold back doing it indeed. On Mac, if I have multiple images open with port 8080 being used, there is no problem.

However, the 'getting started' instructions are way to complicated if people need to understand what an adaptor is, etc. I wanted it to become easier again to just install and browse... This is one of the most asked questions when people are loading Seaside... they don't know what to do next, so we provided them with the magic statement `ZnZincServerAdaptor startOn: 8080`. 

Can we check and/or do something to prevent crashing? If not, then the magic statement will have to do.

jecisc:
I'll try on my Window 7 to see if it still happens.

jecisc:
Ok, now I do not have a crash anymore. Maybe since 2015 something changed in Windows socket management.

jbrichau:
> Ok, now I do not have a crash anymore. Maybe since 2015 something changed in Windows socket management.

Or Pharo's :)

jbrichau:
Unfortunately, it's not working yet since the package is loaded before the Zinc adaptor. I'm looking at fixing this.

Issue #1077, opened 2018-12-17
Possibly remove WAPackagingTest


theseion:
I'm not familiar with the mechanism behind GRPackage but WAPackagingTest doesn't make any sense in its current form.

marschall:
GRPackage has a way to track dependencies, this predates Metacello and over the years has no longer been properly maintained tested.
I think it would make sense the remove `WAPackagingTest` and remove the dependency tracking in `GRPackage` and have Metacello as our primary source of dependency tracking.

theseion:
This issue is on hold. As stated in https://github.com/SeasideSt/Grease/pull/72, we'll investigate what parts of GRPackage are still in use.

jbrichau:
Removing the WAPackagingTest

PR #1100, opened 2018-12-23
Removed extensions to GRPackage

theseion:
refs #1077


theseion:
This PR is related to https://github.com/SeasideSt/Grease/pull/72.

jecisc:
@theseion Should we keep this PR after the comment on the equivalent PR of Grease?

theseion:
Good point, thanks. Closing it.

Issue #1105, opened 2018-12-24
World menu broken in Pharo < 7 because of Seaside


jecisc:
Apparently, the way to display the control panel icon does not work in Pharo < 7 and it breaks the world menu.

To reproduce, just load Seaside in Pharo 6.1 and open the menu.

PR #1106, opened 2018-12-26
Fix crashing adaptor icon reading in Pharo 5 and 6

jbrichau:
Fix for #1105


jecisc:
Tested in P7, no regression. I'll now test in P6

Issue #88, opened 2015-03-25
Updated ColorValue>>fromString: to support 3-digit notation


GoogleCodeExporter:
```
I had just updated fromString method to support three-digit notation and
pass the following test case, was thinking it might be useful to
someone, but again, can't quite say if this is something Cincom folks
need to integrate in their branch of Squeak-Extensions or something that
should go into Squeak's version, so don't jump on my case for
cross-posting please ;)

testColorFromString

 self assert: (ColorValue fromString: 'ff0000') = ColorValue red.
 self assert: (ColorValue fromString: 'FF0000') = ColorValue red.
 self assert: (ColorValue fromString: '#ff0000') = ColorValue red.
 self assert: (ColorValue fromString: '#FF0000') = ColorValue red.
 self assert: (ColorValue fromString: 'f00') = ColorValue red.
 self assert: (ColorValue fromString: 'F00') = ColorValue red.
 self assert: (ColorValue fromString: '#f00') = ColorValue red.
 self assert: (ColorValue fromString: '#F00') = ColorValue red.
 self assert: (ColorValue fromString: 'red') = ColorValue red.
 self assert: (ColorValue fromString: 'RED') = ColorValue red.
 self assert: (ColorValue fromString: '#red') = ColorValue red.
 self assert: (ColorValue fromString: '#RED') = ColorValue red.
```

Original issue reported on code.google.com by `philippe...@gmail.com` on 5 Jul 2008 at 7:56

Attachments:
- [ColorValue class-fromString.st](https://storage.googleapis.com/google-code-attachments/seaside/issue-88/comment-0/ColorValue class-fromString.st)

PR #1111, opened 2019-02-09
Issue #88: add 3.5.0 to baseline ... local tests are passing

dalehenrich:
[All of the GemStone tests passed](https://travis-ci.org/SeasideSt/Seaside/builds/490810666) and the only changes I've made is to the `baselineGemstone` method .. all of the Pharo builds failed `WANoDuplicateUuidsTest` so this must be a pharo thing ... I will go ahead a merge when travis is done, since I am hoping to use this for internal test runs before 3.5.0 is released ...

### DIFF
--- a/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinegemstone..st
+++ b/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinegemstone..st
@@ -97,7 +97,7 @@ baselinegemstone: spec
         package: 'Seaside-GemStone-Core'
           with: [ spec requires: #('Seaside-GemStone300-Core') ] ].
   spec
-    for: #(#'gs3.2.x' #'gs3.3.x' #'gs3.4.x')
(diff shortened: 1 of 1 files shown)


dalehenrich:
SeasideSt/Greass#88 is the issue that is being referred to ...

jbrichau:
@dalehenrich The problem is (was) an upstream change in Grease that breaks the GRPackage dependency descriptions. These are only tested in Seaside. We're probably going to remove these as they basically pre-date the Metacello baseline.... 

Next time, give me a ping before merging a failing PR. I just discovered by accident that the master build was failing. I would want to avoid that.

dalehenrich:
@jbrichau sorry about that, normally I would ping you, but I did need the PR for internal testing (I was doing internal test development and could not make progress without being able to test the master branch of Seaside against 3.5.0) .. and presumably anyone building Pharo Seaside would run into the exact same problems, since I do assume that my changes had nothing to do with the failure .. in a normal universe I would leave it up to you do a merge

PR #1116, opened 2019-03-15
fix for Issue #1115

### LINKED ISSUE #1115: (3.3.2) The message #rfc2965String is sent but not implemented in GemStone 3.4.3
Just did a local test run where I check for sent but not implemented as well as undefined globals and it turned up the fact that #rfc2965String is sent but not implemented ... probably should add the `sent but not implemented` and `undefined globals` tests to at least the GemStone tests while we're addressing this missing method ...

dalehenrich:
fix #1115.

Take a look at [this comment](https://github.com/SeasideSt/Seaside/issues/1115#issuecomment-473408429) where I explain the rationale for the changes.

### DIFF
--- a/.smalltalk.ston
+++ b/.smalltalk.ston
@@ -28,6 +28,8 @@ SmalltalkCISpec {
   #testing : {
       #exclude : {
         #packages : [ 'Cryptography.*' ]
-      }
-  }
+      },
+			#include : {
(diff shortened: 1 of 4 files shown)


dalehenrich:
Squeak builds failing because http://www.squeaksource.com/ is getting a `Bad Gateway` error...

PR #1120, opened 2019-04-13
Add catalog configuration for Pharo

jbrichau:
I updated the ConfigurationOfSeaside3 class for the Pharo7 catalog browser, so it loads from github at least. We should also keep the catalog ConfigurationOfSeaside3 here in the repository so we are reminded about pushing updates to the catalog browser as well.

Can we make a more generic one so we do not need to push updates at all to the MC repo for the catalog browser?


jbrichau:
Hi @jecisc,
I added it as a requirement to approve PR's that at least one other person reviews the request. I can turn that off again or just ignore it, but I think it would be good practice. Let me know what you think :) I added you as a reviewer on this one a couple of weeks ago. I'm happy to not do that again if you want. No sweat :)

Issue #1126, opened 2019-06-02
Seaside Baseline should reference Grease using the semantic tag instead of master head


jbrichau:
On github, we have followed the convention to reference Grease master HEAD, as we keep Grease backwards compatible. However, there have been discussions on discord and in some emails to use semantic versioning tags (e.g. `v1.4.x`), like we used to do on Smalltalkhub with symbolic versions in Metacello (e.g. `#release14x`).

We added these 'semantic versioning tags' to both Grease and Seaside projects. This issue now suggests to use them in the `BaselineOfSeaside3` to reference the Grease version.

PR #1127, opened 2019-06-02
BaselineOfSeaside3 should reference Grease semantic version tag

jbrichau:
See issue #1126


jbrichau:
I messed with Iceberg and worked from another branch than master... deleting this and will re-recreate.

PR #1128, opened 2019-06-02
Reference Grease v1.4.x

jbrichau:
See #1126

### DIFF
--- a/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinecommon..st
+++ b/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinecommon..st
@@ -7,7 +7,7 @@ baselinecommon: spec
         	with: [ 
 				spec
 					loads: #('Grease-Core');
-					repository: 'github://SeasideSt/Grease:master/repository' ].
+					repository: 'github://SeasideSt/Grease:v1.4.x/repository' ].
 		spec
 			project: 'Grease Core Tests' copyFrom: 'Grease' with: [ 
 				spec loads: #('Core Tests') ] 


jecisc:
Travis is green, the change seems good to me.

jvalteren:
Guys, since this change, I'm getting a `Revspec 'v1.4.x' not found` error when loading my own project baseline with a dependency on baseline Seaside3 (at github://SeasideSt/Seaside:master/repository).
When I debug and move waaaay down the stack (am I the only one finding Metacello/Iceberg stacks impossible to read?), I can find the following signal message text:

> You already have an Iceberg repository 
at /Users/jvalteren/Documents/Pharo/images/p7-papertrail/pharo-local/iceberg/SeasideSt/Grease.

What can cause this? I haven't specified any dependencies on Grease.

jbrichau:
Not sure, but I'm guessing your local clone of Grease did not fetch the new tag. Do you get this on a new full load?

jvalteren:
Forgive me, for I was wrong ;-) It turned out that it was indeed having a repository conflict on the local filesystem.

It looks like I have uncovered a severe bug in the Pharo launcher. It seems to set the working of any new image to directory of a previous image, which effectively makes Iceberg share its repository location (because the 'pharo-local' of the other image is being used as repository location).

I will create an issue at the pharo-launcher project.

jvalteren:
PS Grease now does show a `Detached HEAD` status under Iceberg repositories...

jecisc:
It’s normal, because you have checkout a tag and not a branch. So you have a working version of grease, but if you wish to contribute to it, you need to checkout (or create) a branch first. 

For your comment on the share local directory, it might not be a problem of the launcher but of the settings. If you have exported all Pharo settings, it includes a fix location of a local directory, so all images launched with settings will point to this directory.

Issue #1129, opened 2019-06-03
Zinc-Seaside breaks on ZnUrl pathSegment containing `$/`


jbrichau:
The change in Zinc https://github.com/svenvc/zinc/commit/d962746ae9688550d0c826b9103018a87789afb1 breaks an assumption in Zinc-Seaside because pathSegments now can contain single chars (which is the case for a `$/` in the pathSegments.

jbrichau:
The problem is this line of code: https://github.com/SeasideSt/Seaside/blob/a9b91bbc55fbfdb016427bc39c6bb2385009cb81/repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/requestUrlFor..st#L12

noha:
#/ needs to be changed to $/

jbrichau:
@noha Yes, I'm on it ;) But it takes more time to write a test for it than fix the bug itself.

marschall:
I wonder whether something like this would be easier to understand and less error prone

```
	zincUrl hasPath ifTrue: [
		zincUrl isDirectoryPath
			ifTrue: [
				"last path segment is $/ or #/"
				zincUrl pathSegments allButLastDo: [ :each |
					seasideUrl addToPath: each ].
				seasideUrl slash: true ]
			ifFalse: [
				seasideUrl addAllToPath: zincUrl pathSegments ] ].
```
if we don't care about the copy we could even use something like this

```
	zincUrl hasPath ifTrue: [
		| segments |
		segments := zincUrl isDirectoryPath
			ifTrue: [ zincUrl pathSegments allButLast "last path segment is $/ or #/" ]
			ifFalse: [ zincUrl pathSegments ].
		seasideUrl addAllToPath: segments.
		seasideUrl slash: zincUrl isDirectoryPath ].
```

PR #1130, opened 2019-06-03
Fix for issue #1129

jbrichau:
Deal with ZnUrl pathSegments change in [1] where instances of Charact…er can be part of the path segments. Keep backwards compatibility with previous Zinc versions as well by converting to strings when creating a WAUrl instance.

[1] https://github.com/svenvc/zinc/commit/d962746ae9688550d0c826b9103018a87789afb1


noha:
I guess this won‘t work with deliberate slashes in a url aka %2F The decoded slash will be stored as #/ in a segment. Only the trailing slash becomes $/

jbrichau:
@noha I just adapted the existing code to work again in the presence of Character instances because I anticipate this error will soon occur when people load latest Zinc (or install Seaside).

Not sure if we need to cover additional cases in there. For that, I'm not sure why we were filtering them out anyway (yet).

jbrichau:
@marschall I put you as reviewer because I wanted some of your thoughts on why we explicitly filter the trailing slash out. If you do not have the time, no worries, then I merge this PR asap because current Seaside breaks with latest Zinc.

marschall:
I wonder whether something like this would be easier to understand and less error prone

```
	zincUrl hasPath ifTrue: [
		zincUrl isDirectoryPath
			ifTrue: [
				"last path segment is $/ or #/"
				zincUrl pathSegments allButLastDo: [ :each |
					seasideUrl addToPath: each ].
				seasideUrl slash: true ]
			ifFalse: [
				seasideUrl addAllToPath: zincUrl pathSegments ] ].
```
if we don't care about the copy we could even use something like this

```
	zincUrl hasPath ifTrue: [
		| segments |
		segments := zincUrl isDirectoryPath
			ifTrue: [ zincUrl pathSegments allButLast "last path segment is $/ or #/" ]
			ifFalse: [ zincUrl pathSegments ].
		seasideUrl addAllToPath: segments.
		seasideUrl slash: zincUrl isDirectoryPath ].
```

PR #1136, opened 2019-06-22
WaBuilder should be able to render on arbitrary stream

theseion:


jecisc:
@jbrichau Should this go in develop or master?

When you want to switch to a master only contributions, it would be good to update the contribution guide.

jbrichau:
@jecisc Let's keep it on develop. I can check to merge develop to master after verifying what is actually pending.

PR #1140, opened 2019-06-27
Merge develop in master

jbrichau:
Let's get develop branch out of the way and see if we can get all pending changes into a new master release


jecisc:
@jbrichau Why did you revert the change on the icons?

jbrichau:
@jecisc The implementation that is on master is the correct one, using Grease. Somehow, git does not see the following change on master as 'newer':
https://github.com/SeasideSt/Seaside/commit/ce851078886b9c44bede82b61071f8ec6e47061d#diff-569e9b15a6ba6e4db75eb52e9ac93f33

jecisc:
Oh, I got a brain log I read the diff in the wrong way. I agree that the one using grease is the right one!

PR #1149, opened 2019-08-25
1146 Fix WAComboResponse class comment

### LINKED ISSUE #1146: WAComboResponse isn't being flushed as stated in the class comment
The class comment says that `aCanvas flush` will flush the response. That is not correct. Currently we'd need to do something like `WACurrentRequestContext value response flush`.

marschall:
Fixes #1146


pdebruic:
shouldn't this be `self requestContext response flush` ?

theseion:
Looks like it :) Care to open a PR?

pdebruic:
See #1316

PR #1155, opened 2019-08-26
New-control-panel-in-spec2

jecisc:
Introduce a new Control panel in Spec 2 for Pharo 8 +


theseion:
Maybe you want to include #1080?

jecisc:
This is weird because I have the icon and I don't see what might be the problem :(

theseion:
The icon isn't being properly registered. The method on `ThemeIcons` is never executed.

jecisc:
Should be fixes

marschall:
This causes the Squeak builds to fail because we send `SystemVersion>>major`

repository/BaselineOfSeaside3.package/GRPlatform.extension/instance/pharoVersionsFrom..st
```diff
@@ -0,0 +1,5 @@
+*BaselineOfSeaside3
+pharoVersionsFrom: aVersion
+	"Return a list of pharo version from the major version as parameter to the major version of the image"
+
+	^ {}
```
marschall: That's non-platform specific code. We should stick to standard Smalltalk Syntax here and avoid Squeak extensions.

jbrichau:
I'm trying to recover these changes and get them into master. I never investigated in detail but always thought the PR was blocked with requested changes... I have the impression they have been resolved...?

repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinepharo..st
```diff
@@ -49,9 +49,16 @@ baselinepharo: spec
...
+	spec
+		for: #(#'pharo4.x' #'pharo5.x' #'pharo6.x' #'pharo7.x')
+		do: [ spec
 				package: 'Seaside-Tools-Core' with: [ spec includes: #('Seaside-Pharo-Tools-Spec') ];
-				package: 'Seaside-Pharo-Tools-Spec' with: [ spec requires: #('Seaside-Tools-Core') ] ]
\ No newline at end of file
+				package: 'Seaside-Pharo-Tools-Spec' with: [ spec requires: #('Seaside-Tools-Core') ] ].
+
+	spec
+		for: (GRPlatform current pharoVersionsFrom: 8)
```
jbrichau, line 61: shouldn't this be `self pharoVersionsFrom: 8` ? 
(not sure if I like this way, but for sure it's where the travis builds fail on...)

jbrichau: @jecisc if possible, could you just change it to `pharo8`? I prefer to manually update for new versions rather than trying to figure out later on if some package should have a `self pharoVersionsBetween: xx and: yy` method :)

jbrichau:
I forgot I can very easily complete this myself by merging it into another branch first and make the necessary changes. Going ahead with that.

jecisc:
Hi @jbrichau 
Thank you for taking care of that! Was a little busy this week end and you were faster than me to update the baseline :)

For info: When I do a PR I let the maintainers of the projects edit my PR. This means that you can checkout my branch, change it and push directly on the branch in my fork and it will update the current PR without the need of creating a new one.

jbrichau:
@jecisc I did not try to see if I could write to your branch, so the easiest for me was to quickly pull it into another branch.
There still is an issue though: https://github.com/SeasideSt/Seaside/pull/1214#issuecomment-657197587

I don't know if you have seen that before and what the best way to prevent it is?

(it's the same issue as raised by Max before)

PR #1158, opened 2019-08-26
1157 validate percent values are hex

### LINKED ISSUE #1157: URL parsing should validate percent are hex
We should validate percent values are valid hex, for example `%ZZ` should not be valid.

marschall:
Fixes #1157

### DIFF
--- a/repository/Seaside-Core.package/WAUrl.class/class/decodePercent..st
+++ b/repository/Seaside-Core.package/WAUrl.class/class/decodePercent..st
@@ -12,8 +12,8 @@ decodePercent: aString
 				char = $%
 					ifTrue: [
 						| firstByte secondByte |
-						firstByte :=  input atEnd ifTrue: [ WAInvalidUrlSyntaxError signal: aString ] ifFalse: [ input next asUppercase digitValue ].
-						secondByte :=  input atEnd ifTrue: [ WAInvalidUrlSyntaxError signal: aString ] ifFalse: [ input next asUppercase digitValue ].
-						Character codePoint:firstByte * 16 + secondByte ]
+						firstByte :=  self readHexFrom: input errorDetail: aString.
+						secondByte :=  self readHexFrom: input errorDetail: aString.
+						Character codePoint: firstByte * 16 + secondByte ]
 					ifFalse: [ char ] ]) ].
 	^ output contents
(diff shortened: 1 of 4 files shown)


repository/Seaside-Core.package/WAUrl.class/class/decodePercent..st
```diff
@@ -12,8 +12,8 @@ decodePercent: aString
 				char = $%
 					ifTrue: [
 						| firstByte secondByte |
-						firstByte :=  input atEnd ifTrue: [ WAInvalidUrlSyntaxError signal: aString ] ifFalse: [ input next asUppercase digitValue ].
-						secondByte :=  input atEnd ifTrue: [ WAInvalidUrlSyntaxError signal: aString ] ifFalse: [ input next asUppercase digitValue ].
+						firstByte :=  self readHexFrom: input errorDetail: aString.
+						secondByte :=  self readHexFrom: input errorDetail: aString.
 						Character codePoint:firstByte * 16 + secondByte ]
```
theseion: Missing space

repository/Seaside-Core.package/WAUrl.class/class/readHexFrom.errorDetail..st
```diff
@@ -0,0 +1,10 @@
+private
+readHexFrom: aStream errorDetail: aString
+	| hexChar |
+	aStream atEnd ifTrue: [
+		WAInvalidUrlSyntaxError signal: aString ].
+	hexChar := aStream next.
```
theseion: Would bit shift operations make sense here to improve performance?

marschall: The temptation to optimize is there but I think it's unlikely this shows up during profiling.

marianopeck:
Hi @marschall @theseion @gcotelli 

I was reviewing this old issue and I see you modified `#decodePercent:` to delegate to ` #readHexFrom:errorDetail:`
My question is whether we should actually do the same for `#decodePercent:from:to:` and `#decodePercentAndPlus:from:to:`

thanks in advance

gcotelli:
IMHO yes. Probably nobody noticed the problem with the other methods. Do you want to provide a PR and add some test cases? Looks that both methods are used during parsing.

theseion:
I agree.

Issue #1145, opened 2019-07-31
WAComboResponse doesn't work with socket streams


theseion:
ZdcSocketStream (Zinc) does not implement `#crlf` and `stream nextPut: Character space` fails because Characters can no longer be stored into a byte array (`#nextPutAll:` works because of the implicit conversion done by the VM in the primitive).

theseion:
These issues would be taken care of by https://github.com/SeasideSt/Grease/pull/138, which simply introduces two extensions. But see discussion in #1159. This might not be the best solution.

PR #1159, opened 2019-08-26
POC implementation of WAComboResponse acting directly on a socket stream

theseion:


theseion:
I also added this method to GRPlatform: 
```
doesVmSupportStringToByteArrayConversion
	^ doesVmSupportStringToByteArrayConversion ifNil: [
		doesVmSupportStringToByteArrayConversion := [
			(ByteArray new: 1)
				replaceFrom: 1
				to: 1
				with: 'a'.
			true ]
					on: Error
					do: [ false ] ]
```

theseion:
See #1145

marschall:
> I also added this method to GRPlatform:
> 
> ```
> doesVmSupportStringToByteArrayConversion
> 	"..."
> ```

I assume this method would be used to decide whether we need to wrap in a `BinaryConvertingStream`. I wonder whether we could do without this and instead let either a `GRPlatform` method or adapter method return the correct type of stream.

> And `WABinaryConvertingStream`.

I assume you mean `GRBinaryConvertingStream`

theseion:
@marschall As far as I can tell there are no facilities on GRPlatform yet that deal with differences in VM's, which might be an issue in this case, when there are multiple VM's to choose from for a given platform.
Where would you put the test on the adapter? The problem I see there is that we'd have to pass the adaptor as an argument to the response object that we're creating (see `ZnZincStreamingServerAdapter>>responseFor:`), which feels wrong to me. It's not a property of the chosen server.

I do agree that it doesn't feel quite right on GRPlatform either. Maybe we need to add something like `GRVmFacilities`?

jbrichau:
@theseion I removed the develop branch but did not intend to discard this PR. I know you're on other tracks right now, but would you mind pulling this change against master? I don't know the status and if someone can finish it, but would be good to keep track of it. Thanks!

PR #1160, opened 2019-08-27
Remove code duplication and fix remoteaddress-based tracking strategy and filter

jbrichau:
Remove code duplication and fix for retrieving value in forwarded header -or- the remoteAddress when relying on the origin ip in the `WARemoteAddressProtectionFilter` and `WAIPSessionTrackingStrategy`

In addition, the`WARemoteAddressProtectionFilter` was storing the value from the forwarded header (if present) but the first comparison was still returning the remoteAddress anyway. So this can never have worked in the case where using the header was necessary.

Fixes for 1152

### DIFF
--- a/repository/Seaside-Core.package/WARemoteAddressProtectionFilter.class/instance/remoteAddressFromRequest..st
+++ b/repository/Seaside-Core.package/WARemoteAddressProtectionFilter.class/instance/remoteAddressFromRequest..st
@@ -1,11 +1,6 @@
 private
 remoteAddressFromRequest: aRequest
-	remoteAddress ifNil: [
-		remoteAddress := aRequest
-			headerAt: 'x-forwarded-for'
-			ifAbsent: [
-				aRequest
-					headerAt: 'forwarded-for'
-					ifAbsent: [
-						aRequest remoteAddress ] ] ].
-	^ aRequest remoteAddress
+	| result |
+	result := aRequest forwarded ifNil:[ aRequest remoteAddress ].
+	remoteAddress ifNil:[ remoteAddress := result ].
+	^ result
--- a/repository/Seaside-Core.package/WARemoteAddressProtectionFilter.class/instance/remoteAddressFromRequest..st
+++ b/repository/Seaside-Core.package/WARemoteAddressProtectionFilter.class/instance/remoteAddressFromRequest..st
@@ -1,6 +1,6 @@
 private
 remoteAddressFromRequest: aRequest
 	| result |
-	result := aRequest forwarded ifNil:[ aRequest remoteAddress ].
+	result := aRequest forwardedFor ifNil:[ aRequest remoteAddress ].
 	remoteAddress ifNil:[ remoteAddress := result ].
 	^ result
--- a/repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
+++ b/repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
@@ -1,0 +1,3 @@
+accessing-headers
+forwarded
+	^ self headerAt: 'forwarded' ifAbsent:[ self headerAt: 'x-forwarded-for' ]
--- a/repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
+++ b/repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
@@ -1,3 +1,3 @@
 accessing-headers
 forwarded
-	^ self headerAt: 'forwarded' ifAbsent:[ self headerAt: 'x-forwarded-for' ]
+	^ self headerAt: 'forwarded'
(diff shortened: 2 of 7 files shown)


repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
```diff
@@ -0,0 +1,3 @@
+accessing-headers
+forwarded
+	^ self headerAt: 'forwarded' ifAbsent:[ self headerAt: 'x-forwarded-for' ]
```
theseion: I don't think this will work for the `Forwarded` header, as the contents would be `for=192.0.2.43, for="[2001:db8:cafe::17]"` or `for=123.34.567.89`. So we need to extract the first occurrence of the `for=` attribute.

repository/Seaside-Core.package/WARemoteAddressProtectionFilter.class/instance/remoteAddressFromRequest..st
```diff
@@ -1,11 +1,6 @@
...
-			ifAbsent: [
-				aRequest
-					headerAt: 'forwarded-for'
-					ifAbsent: [
-						aRequest remoteAddress ] ] ].
-	^ aRequest remoteAddress
\ No newline at end of file
+	| result |
+	result := aRequest forwarded ifNil:[ aRequest remoteAddress ].
+	remoteAddress ifNil:[ remoteAddress := result ].
```
theseion: Why do we need to check `#forwarded` every time? Can it change?

repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
jbrichau: My intention is that this accessor returns the entire header (as a string). We don't try to parse it and create a `WARemoteAddress` object (or something like that). Since none of the uses in the protection filter or the ip-based tracking strategy really require to parse it as an address, but rather only use it as a value to index on or compare, I think that should be sufficient.

repository/Seaside-Core.package/WARemoteAddressProtectionFilter.class/instance/remoteAddressFromRequest..st
jbrichau: Not sure I understand what you mean. We need to check the remoteAddress (or the value in the forwarded header) in every request, no?

theseion: Are you thinking of mobile clients?

jbrichau: not really, the filter just needs to verify the value in every request. It sets the value it extracts from the first request it processes and requires any subsequent request to have the same value.

I do not intend to improve this filter functionally, just fix the code.

marschall:
I am a bit uncomfortable with simply reading the request headers without knowing anything about the setup of the deployment a client could simply send a `Forwarded` header to "fake" his IP.

jbrichau:
Ok, but that was what the filter does. I just tried to fix code duplication and the fact that the header name was not using the current standard name. Should we remove the filter alltogether? I'm not against it... but we might want to just mark that fact in the comments...

jbrichau:
I added the parsing of the 'For' directive to use the identifier instead of the entire header (which was wrong). So, now let's see who dares to use an ip-based tracking strategy or protection-filter :)

repository/Seaside-Core.package/WARequest.class/instance/forwarded.st
```diff
@@ -1,3 +1,3 @@
 accessing-headers
 forwarded
-	^ self headerAt: 'forwarded' ifAbsent:[ self headerAt: 'x-forwarded-for' ]
\ No newline at end of file
+	^ self headerAt: 'forwarded'
```
theseion: Shouldn't we keep the fallback to `x-forwarded-for`?

jbrichau: It had a different syntax. I would say we do not support it.

theseion: Ok. Given that the code has never worked in the first place we can do that. But will that work for you when you upgrade to 3.4 from 3.0? If I'm not mistaken your using this at work, no?

jbrichau: Not the ip-based one. I just had to rename that class which unleashed the changes in this PR ;)

jbrichau: We use the cookie-based protection filter I added

PR #1161, opened 2019-08-27
Added missing super send to JSAlias>>javascriptContentOn:

### LINKED ISSUE #1057: Ajax code is not generated after javascript function call
I have a button with an Ajax call and the call of a javascript function. Depending on the order of the statements, it works or it does not work.

For reproduction  I rendered two buttons 

```
html div
     onClick:
            (html jQuery ajax script: [ :script | #script inspect ])
                ,
                    ((html javascript alias: 'getS3Data')
                        apply: {((html jQuery ajax callback: [ :var | #var inspect ] 
                        value: (html javascript alias: 'value')) asFunction: #(value))});
        with: 'works'.
html div
        onClick:
            ((html javascript alias: 'getS3Data')
                        apply: {((html jQuery ajax callback: [ :var | #var inspect ] 
                        value: (html javascript alias: 'value')) asFunction: #(value))}),
                        (html jQuery ajax script: [ :script | #script inspect ]);
        with: 'does not work'.
```

It does not render the ajax in the 2nd case:

```
<div onclick="$.ajax({...});getS3Data(function(value){$.ajax({...})})">works</div>
<div onclick="getS3Data(function(value){$.ajax({...})})">does not work</div>
```
the javascript function does some s3 script and returns the callback value like this:

```
<script>		
function getS3Data(callback) \{
...
		callback(content);
	...
} 
</script>	
```	

For completeness, screenshot of the rendered result:

bildschirmfoto 2018-11-28 um 09 10 46

theseion:
Fixes #1057


marschall:
A test would be nice :smiley:
Maybe the what Sabine reported can be incorporated in a regression test in `JQAjaxTest`.

theseion:
Yes, I'll do that.

theseion:
Added test.

jbrichau:
I think this is only a part of the solution. The following still breaks.
Try:
```
	| script |
	WAHtmlCanvas builder
		render: [ :html | 
			script := (html javascript alias: 'externalexpression'), 
						 ((html javascript alias: 'internalexpression') asFunction: #(value)) ].
					
	self assert: script equals: 'externalexpression;function(value){internalexpression}'.
		
	WAHtmlCanvas builder
		render: [ :html | 
			script := ((html javascript alias: 'internalexpression') asFunction: #(value)),
						 (html javascript alias: 'externalexpression') ].
	self assert: script equals: 'function(value){internalexpression};externalexpression'
```

jbrichau:
I've been diving into this (again) and the way `JSScript` can be composed is just strange and buggy. It might be that it was never intended to write these kinds of compositions, but it seems natural to be able to do so, so I'm trying to find a good solution.

The current change breaks other tests (and behavior) unfortunately.

PR #1163, opened 2019-08-27
Also test on Squeak 5.2

marschall:


jbrichau:
I forgot: can you also add the Squeak 5.2 badge to the README? 
We added those because people were asking, so best keep them updated.

PR #1164, opened 2019-08-27
Revert allowing Squeak 5.2 failures

marschall:
My previous commit was not good. We already test on Squeak 5.2,
the tests are green and it is even in the badges.


jbrichau:
Ha.. it's too hot to concentrate ;)

PR #1189, opened 2019-10-27
Add dynamic mime document in tests

### LINKED ISSUE #1188: Add example on how to dynamically generate resources
http://lists.squeakfoundation.org/pipermail/seaside-dev/2019-October/007251.html

marschall:
Add dynamic mime document in tests as an example on how to dynamically generate resources.

Closes #1188

### DIFF
--- a/repository/Seaside-Tests-Core.package/WABlockMimeDocument.class/README.md
+++ b/repository/Seaside-Tests-Core.package/WABlockMimeDocument.class/README.md
@@ -1,0 +1,13 @@
+I am a MIME document whos contents are generated dynamically by a block.
+
+I can be used like this:
+
+html image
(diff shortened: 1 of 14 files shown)


marschall:
A couple of implementation notes:

* `WAMimeDocument` should have an abstract base class that only defines the protocol like the class comment suggests. I can do this in a different PR. Defining `#=` is hard, the easiest is to simply leave it out.
* overriding `#seasideMimeDocument` and `#seasideMimeDocumentType:` on `BlockClosure` would be the easiest but would be platform specific.
* It can be argued that we should implement the same class side `#on:(mimeType:(fileName:)) `protocol, however the first argument will have a different type.

PR #1194, opened 2020-05-16
Fix forgotten seaside version increase in WAWelcome

jbrichau:


jbrichau:
I merge this anyway because the build only fails because of Squeak download issue and the change is minimal.

Issue #1198, opened 2020-05-21
WAFlowFunctionalTest fails on GemStone


dassi:
### Bug summary

`WAFlowFunctionalTest` fails: Exception `WARequestContextNotFound` after seaside's call/answer

### Steps

1. Use GsDevKit, install GemStone 3.4.5 (probably not specific to that version, also failed with GemStone 3.4.3).
2. Install Seaside with the "Tests" group, which includes the functional tests.
3. Browse to your domain http://localhost/tests/functional/WAFlowFunctionalTest
4. Both subtests fail with a `WARequestContextNotFound`

That simply means, that seaside's very basic call:/answer is not working.

### My observations

- Bug DOES NOT show up using `call:onAnswer:` instead of `call:`.
- Bug is NOT there up to Seaside v3.2.5. But is there starting from Seaside v3.3.0. Grease has done a jump there, changing the platform specific implementation of `WADynamicVariable` (which `WACurrentRequestContext` is a subclass of). On GemStone this is now a subclass of GemStone's own `DynamicVariable` class, which in turn uses `Processor activeProcess environment at:` to get the value of the dynamic variable. I managed to figure out, that the "Processor activeProcess" changes after call:/answer, which could be OK but the process' environment dictionary is blank and the current request context is gone.
- Seaside's error handling will in turn fail itself because it also relies on calling current requestContext itself. Only a simple text based stack is placed on GemStone's ObjectLog. And the browser just shows a simple "Internal Error:" text. Not easy to debug.

dassi:
@jbrichau suggested:
> In the meantime, you might want to change WADynamicVariable to be a subclass of GRNotificationBasedDynamicVariable instead of GRDynamicVariable and see if that fixes it?

Yeeeaahhh! That fixes it!

dalehenrich:
@dassi Glad that works for you.

> Seaside's error handling will in turn fail itself because it also relies on calling current requestContext itself. Only a simple text based stack is placed on GemStone's ObjectLog. And the browser just shows a simple "Internal Error:" text. Not easy to debug.

@dassi I noticed this as well, but when the continuation mechanism fails, I'm not sure there is much to do...

dalehenrich:
@jbrichau,

From this stack:
```
1 GRGemStonePlatform >> logError:title:shouldCommit: @3 line 4  [GsNMethod 144132865]
2 GRGemStonePlatform >> logError:title: @2 line 3  [GsNMethod 144184321]
[...]

PR #1201, opened 2020-05-24
Fix for issue 1198: Restore of partial continuations does not preserve GsProcess environment, breaking dynamic variables

jbrichau:
See discussion in #1198


jbrichau:
I'll do a squask and merge when this PR gets merged eventually (hint: don't look at the commits, just at the 'files changed' in this PR 🙄

jbrichau:
Not entirely finished as some more tests break (obviously) because of this workaround patch. Will continue later...

jbrichau:
Failing Squeak and Gemstone3.5 builds are unrelated to the changes in this PR.
Need to remove some monticello metadata that popped up when publishing from GS and then this should be good to go.

Issue #1204, opened 2020-06-18
WAPharoWalkback is not suitable for use in Squeak


timrowledge:
The method as-is

```
WAPharoWalkback >> tempNamesAndValuesIn: aContext do: aTwoArgumentBlock 
	SystemVersion current major >= 8
		ifTrue: [aContext tempNames
				doWithIndex: [:each :index | aTwoArgumentBlock
						value: each
						value: (aContext namedTempAt: index)]]
```
cannot work in Squeak since
a) SystemVersion uses #majorVersionNumber rather than #major (seems ot have been the case since 2010 at least)
b) needs to check for  >5 not >8
c) maybe doesn't need the check at all?

So a workable version could be 
```
tempNamesAndValuesIn: aContext do: aTwoArgumentBlock 
	SystemVersion current majorVersionNumber >= 5
		ifTrue: [aContext tempNames
				doWithIndex: [:each :index | aTwoArgumentBlock
						value: each
						value: (aContext namedTempAt: index)]]
```
I imagine the simple work-around would be a subclass of WAPharoWalkback 'WQSqueakWalkback' placed in a Seaside-Squeak-Development package?

krono:
What about the following, which also restores pre-Pharo8 compat?

```Smalltalk
WAPharoWalkback >> tempNamesAndValuesIn: aContext do: aTwoArgumentBlock 
	(aContext respondsTo: #tempNames)
		ifTrue: [aContext tempNames doWithIndex:
			[:each :index | aTwoArgumentBlock
				value: each
				value: (aContext namedTempAt: index)]]
		ifFalse: [aContext tempScopedNames keysAndValuesDo:
			[:index :each | aTwoArgumentBlock
				value: each
				value: (aContext tempAt: index)]]
```

jbrichau:
Thanks for the suggestion @krono but the methodname is only part of the problem. See the original issue https://github.com/SeasideSt/Seaside/issues/689

Obviously, given the name `WAPharoWalkback`, I was not expecting this is being used in Squeak.... but it seems the package is indeed shared with Squeak.

PR #1209, opened 2020-07-03
Fix WAPharoCallback for use in Squeak

jbrichau:
Fix for issue #1204


jbrichau:
That concern was on my mind indeed. Creating a separate package for it would also mean we need to maintain it as a separate package in the future. Since I'm not using Squeak and nobody is actively maintaining it for Squeak other than us, I opted for the current change.

thx for the review!

Issue #15, opened 2015-03-25
Memory tool mis-sorts total size


GoogleCodeExporter:
```
The memory tool ("Memory" in the toolbar) mis-sorts the "Total Size" column 
when the user clicks 
on the headers. 

WAMemory>>buildTable seems to be the source of the problem. The column contains 
positive 
integers, but they're rendered as strings with commas ("12,345"); sorting those 
in string order 
doesn't produce anything useful.

```

Original issue reported on code.google.com by `renggli` on 2 Mar 2008 at 8:28

Issue #10, opened 2015-03-25
SUNavigation should only report #selected tab (SUTabPanel) or #visible accordian panel (SUAccordion) in #children


GoogleCodeExporter:
```
It currently includes all its elements, including those that are not rendered.

```

Original issue reported on code.google.com by `renggli` on 2 Mar 2008 at 8:25

PR #1210, opened 2020-07-09
Improving WARequestTest >> testPostFields

mabdi:
Hi,

I submit this pull request to suggest some improvements in the test method #testPostFields to cover some corner cases in the class WARequest.

By including the lines #5-#10 the test will now execute methods in the class under test which previously were not covered by WARequestTest (isXmlHttpRequest, headers, remoteAddress, isPost, sslSessionId). This will also guard against a WARRequest object which is improperly initialized.
Line #6 verifies that the WARRequest does not have an XMLHttpRequest header by default.
Line #9 on the other hand verifies that WARequest is properly initialized via #method:uri: and ‘POST’ argument.

By including Line #15, the test will guard against sending #bodyDecoded when charSet isNil (a WAMimeType(application/x-www-form-urlencoded) has a nil charSet).

Note that these suggestions are adapted from a test amplification tool called SmallAmp (https://github.com/mabdi/small-amp). SmallAmp executes existing tests, sees which parts of the class under test are not covered and then suggests improvements on the test methods.

I hope you will accept this pull request. It would illustrate that SmallAmp makes relevant suggestions.

Mehrdad Abdi.

### DIFF
--- a/repository/Seaside-Tests-Core.package/WARequestTest.class/instance/testPostFields.st
+++ b/repository/Seaside-Tests-Core.package/WARequestTest.class/instance/testPostFields.st
@@ -2,10 +2,17 @@ tests
 testPostFields
 	| request headers |
 	request := WARequest method: 'POST' uri: '/foo?bar=1'.
+	self
+		deny: request isXmlHttpRequest;
+		assert: request headers class equals: WAHeaderFields;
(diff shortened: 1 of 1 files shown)


mabdi:
I am closing this pull-request because I believe it's on the wrong branch. I will resend it to the 'master' branch.

PR #1212, opened 2020-07-10
Fixes  to multiple value headers convertion #1211

### LINKED ISSUE #1211: Multiple value Headers only set the latest one when converting a WAResponse to ZnResponse
In `ZnZincServerAdaptor>>#responseFrom:` the header conversion from `WAHeaderFields` to `ZnHeaders` if a header appears several times in `aWAHeaderFields` only the last one in the iteration is added to `ZnResponse` headers.

```smalltalk
"<snip>"
partialHeaders	keysAndValuesDo: [ :key :value | fullHeaders at: key put: value greaseString ].
```

It does `at:put ` and it should do `at:add:` instead.

eMaringolo:
Fixes  #1211

### DIFF
--- a/repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/responseFrom..st
+++ b/repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/responseFrom..st
@@ -6,7 +6,7 @@ responseFrom: aRequestContext
 	cookies := seasideResponse cookies.
 	fullHeaders := ZnHeaders defaultResponseHeaders.
 	partialHeaders keysAndValuesDo: [ :key :value |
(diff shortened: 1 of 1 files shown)


jbrichau:
Hey @eMaringolo 
Thanks for the fix!

I think a small test to prevent regression in the future would be a good idea as well. However, it appears we do not have a test package for the Zinc adaptor, so I registered https://github.com/SeasideSt/Seaside/issues/1213

eMaringolo:
Yes, I noticed that the adaptor had no test case, and that's why I went straight and submitted the proposed fix.

Are these fixes backported somehow to previous releases? (e.g. Seaside 3.3 or 3.2)

PR #1214, opened 2020-07-12
New control panel for Pharo8+ using Spec2

jbrichau:
Recreating the PR of @jecisc (#1155) to finish up some leftovers.


jbrichau:
Looks good except the menu icon still only appears after recompiling the menu method. Would be good to fix that and it probably originates from a load order issue where this method is compiled before the icon is present.

PR #1218, opened 2020-08-05
WAURL>>#urlWithoutSeasideFields

eMaringolo:
Adds #withoutSeasideQueryFields to WAUrl

### DIFF
--- a/repository/Seaside-Core.package/WAUrl.class/instance/withoutSeasideQueryFields.st
+++ b/repository/Seaside-Core.package/WAUrl.class/instance/withoutSeasideQueryFields.st
@@ -1,0 +1,12 @@
+copying
+withoutSeasideQueryFields
+	"Returns a copy of receiver without the DEFAULT query fields
+	used for tracking session (_s), continuation (_k) or document (_d)."
+
+	| newUrl |
+	newUrl := self copy.
+	newUrl queryFields
+		removeKey: '_s' ifAbsent: [  ];
+		removeKey: '_k' ifAbsent: [  ];
+		removeKey: '_d' ifAbsent: [  ].
+	^ newUrl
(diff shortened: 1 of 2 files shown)


jbrichau:
Hi @eMaringolo 
Sorry for the long wait.

The implementation assumes that the session key is `_s` and the continuation key is `_k`. However, they can be overridden and changed in the `keyField` and `actionField` methods. To make this work in the framework, we should use the same parameters, which would require a session to be present in this method implementation.

Would that make sense in your use case scenario?

eMaringolo:
I think I'd delegate it to another version that performs the lookup.

E.g.
```smalltalk
urlWithoutSeasideFields
  "Returns a copy of receiver without Seaside DEFAULT specific query fields,
   (_s for session ,_k for continuation and _d for document)."

  ^self urlWithoutSeasideFields: true
```

```smalltalk
urlWithoutSeasideFields: aBoolean 
  "Returns a copy of receiver without Seaside specific query fields,
   If aBoolean is true it uses the default fields (_s for session ,_k for continuation and _d for document),
   otherwise it delegates the resolution of the field names to the current session."
   ...
```

Another alternative is to pass the list of fields by default (so it behaves similarly to a `copyWithoutAll:`)

I don't like how the keyword selector reads,  but mostly I'd like to avoid having to depend on a WASession, since I'm using it also in contexts where there isn't a session yet (and plan to extend on that use :-) ).

So if this can be adapted to comply with Seaside's coding practices, I can adapt the PR to include such changes (and probably a different selector), otherwise I'll be happy to continue using it as an extension and maybe integrate it back later if what I'm doing takes a larger shape and is worth integrating into the main Seaside codebase.

jbrichau:
I think it might make sense add a ‘copyWithoutQueryFields: aCollection’ method to WAUrl and use that method in your custom extension. I’m being a little picky because the “seaside fields” are configurable and having a method in the framework that is named as such and does not follow that configuration is confusing at least.

jbrichau:
But then I notice the method isSeasideField: https://github.com/SeasideSt/Seaside/blob/master/repository/Seaside-Core.package/WAUrl.class/instance/isSeasideField..st 
Which assumes any query field prefixed with '_' is a Seaside field. So, we can also stick with that :)

eMaringolo:
I wasn't aware of such method. 
Then the `withoutSeasideQueryFields` can be adapted to filter fields using that. I'll adapt the PR to look for that instead of the predefined field names.

Issue #1223, opened 2020-09-03
The class GsContext is in both Seaside (Seaside-GemStone-Core) and Grease (Grease-GemStone-Core)


dalehenrich:
It looks like GsDevKit/Grease does not have GsContext and SeasideSt/Grease does have GsContext.

Since SeasideSt/Grease is going to be the primary Grease repo (finally - GsDevKit/GsdevKit#124) we need to remove the duplication is Seaside-GemStone-Core ...

jbrichau:
I think GsDevKit/Grease is identical to SeasideSt/Grease since some time, no?
The class is there: https://github.com/GsDevKit/Grease/tree/master/repository/Grease-GemStone-Core.package/GsContext.class

It did get moved from Seaside to Grease because of https://github.com/SeasideSt/Grease/issues/40
(which is unfinished but is there since 2 years waiting for me to write a test and hit the close button :)

dalehenrich:
@jbrichau the GLASS1 baseline has been using GsDevKit/Grease ... the switching of the git repository for existing GsDevKit_home installations is the sticking point ... and has gummed up the works for me ... I'm hoping that I will be able to come up with a satisfactory solution while I'm doing the work for 3.6.0, since I have to make changes in a couple of different places ... wish me luck:)

dalehenrich:
@jbrichau ... I just figured out what the REAL issue is ... GLASS1 baseline requires Grease and GLASS1 can be used with both Seaside3.2 (where the change has not been made) and Seaside3.4 where the change HAS been made ...

I will need to think about the best way to resolve this particular issue ... since the change in Grease will impact which Seaside can be used ... sort of as GsContext is basically the same as well .. just moved ...

dalehenrich:
@jbrichau the plot thickens as it looks like [Seaside-GemStone-Core package](https://github.com/SeasideSt/Seaside/tree/69837a66c516e722a9e30cd377c17c5abbbd573f/repository/Seaside-GemStone-Core.package/GsContext.class) still has the GsContext class present ... which means that you get dirty GemStone package "conflicts" when using SeasideSt/Grease ... and this was the issue I was trying to address ...

dalehenrich:
[...]

PR #1227, opened 2020-10-02
Removed the GsContext class because it is in Grease since 2 years now

jbrichau:
Fix for #1223


dalehenrich (dismissed):
Looks good! Thanks!

jbrichau:
@dalehenrich github is not so friendly to a reviewer when new pushes get added :-) I did not dismiss anything. Seems like that happens automatically.

Anyway, a test is failing due to some Duration bug in Pharo.. so I adapted the test to work around that.

dalehenrich (approved):

Issue #1222, opened 2020-09-02
Drop Swazoo from Seaside for GemStone?


dalehenrich:
For 3.6.x support a change needs to be made to the Swazoo2 server code: GsProcess>>terminate will block and ultimately timeout with an error if a socket is waiting on a GsSocket>>accept call ... The code needs to be changed to use GsSocket>>readWillNotBlockWithin: which  will not block on #terminate. It's pretty fundamental for a Seaside server process to be terminated (without errors) and it is relatively straightforward to arrange to insert the use of GsSocket>>readWillNotBlockWithin:. The Zinc-based server code used by GsApplicationTools already uses GsSocket>>readWillNotBlockWithin:.

I was in the process of making this change to the Swazoo code, when I realized that the Swazoo2 project is on SmalltalkHub which is now a static site. Before I go off and create github project for Swazoo2, so I can make the code changes or remove the tests, I thought I would check to see how important it is to continue to support the Swazoo2 code?

I am not aware of anyone actively using Swazoo2 with Seaside. 

The latest configurations are on SmalltalkHub, and there are alternate Monticello repositories (in order of "freshness"): 
- http://seaside.gemtalksystems.com/ss/swazoo2.html
- http://ss3.gemtalksystems.com/ss/Swazoo.html
- http://www.squeaksource.com/Swazoo.html

Clearly no one else was motivated to save the code, so I'm not sure myself if I should make the effort to create a Swazoo github project.

Right now my inclination is to simply remove the project from the Seaside baseline and see if anyone complains. 

If anyone else does care, it is easy enough to add the project back to the baseline and once someone else has established the new repository of record, I will be able to push my changes there ...

What do you guys (@jbrichau and @marschall) think?

jbrichau:
I have no objection to drop support for it. We have Zinc and FastCGI as possibilities.
We can drop Swazoo for Gemstone 3.6 and upwards.

marschall:
I'm fine with removing it as well.

dalehenrich:
Just to be clear which one of the following are you guys (@jbrichau @marschall ) agreeing too?
1. remove Swazoo2 from Seaside for all platforms
2. remove Swazoo2 from Seaside for gemstone 3.6.x and beyond
[...]

PR #1228, opened 2020-10-05
Remove Swazoo project from 3.6 (#1222); complete the fix for Parasol/Seaside infinite loop (SeasideSt/Parasol#43)

dalehenrich:
fix for #1222 .

Once this PR is merged, the [GemStone SeasideSt/Parasol travis tests](https://travis-ci.org/github/SeasideSt/Parasol/jobs/732860827) should start passing ... baseline modifications were needed to ensure that Parasol-Seaside package was loaded **AFTER** Seaside-Tools-Core (where **WAAdmin** is loaded)


dalehenrich:
@jbrichau, Squeak-5.2 seems to be the only case failing the builds and that has been true for some time now ...

dalehenrich:
@jbrichau ... the 3.6.0 release is on a fast track ... I've got two tests that I can work on without this merge, but the next phase of work involves running internal tests against the master branch of Seaside and while I can fiddle things to use this branch, it is much simpler for me to be able to use the master branch ... if you need more time to review, I understand and can wait a couple more days, but then I'll need to decide whether I need to start making changes to the internal tests so I can have time to fix any problems that might show up without delaying the release (I think we're targeting something like 2 weeks out), so the sooner I know about problems the sooner I can go about fixing them:)

I know that I can press the merge button, but requiring that Parasol has to be loaded from Seaside and not the other away around for GemStone could be a problem and I don't want to force this change on you if you don't approve ... again, knowin gsooner rather than later would be a good thing at this stage ...

jbrichau (approved):

dalehenrich:
THANKS, @jbrichau!

PR #1240, opened 2021-02-16
POC implementation of WAComboResponse acting directly on a socket stream

theseion:
This is a clone of #1159, which was opened against the now removed `develop` branch.

### DIFF
--- a/repository/Seaside-Core.package/WAComboResponse.class/instance/commit.st
+++ b/repository/Seaside-Core.package/WAComboResponse.class/instance/commit.st
@@ -8,7 +8,10 @@ commit
 	self headerAt: 'Transfer-Encoding' put: 'chunked'.
 
 	committed := true.
-	self writeStatusOn: externalStream.
-	self writeHeadersOn: externalStream.
-	self writeCookiesOn: externalStream.
-	externalStream crlf; flush
+	self
+		writeStatusOn: externalStream;
+		writeHeadersOn: externalStream;
+		writeCookiesOn: externalStream.
+	externalStream
+		crlf;
+		flush
--- a/repository/Seaside-Core.package/WAComboResponse.class/class/external..st
+++ b/repository/Seaside-Core.package/WAComboResponse.class/class/external..st
@@ -1,5 +1,6 @@
 instance creation
 external: anExternalStream
 	^ self
-		onBuffered: (GRPlatform current writeCharacterStreamOn: (String new: 4096))
-		external: anExternalStream
+		onStream: anExternalStream
+		bufferSize: 4096
+		codec: GRNullCodec new
--- a/repository/Seaside-Core.package/WAComboResponse.class/class/onBuffered.external..st
+++ b/repository/Seaside-Core.package/WAComboResponse.class/class/onBuffered.external..st
@@ -1,3 +1,0 @@
-instance creation
-onBuffered: aBufferedStream external: anExternalStream
-	^ self basicNew initializeOnBuffered: aBufferedStream external: anExternalStream
--- a/repository/Seaside-Core.package/WAComboResponse.class/class/onStream.bufferSize.codec..st
+++ b/repository/Seaside-Core.package/WAComboResponse.class/class/onStream.bufferSize.codec..st
@@ -1,0 +1,5 @@
+instance creation
+onStream: aStream bufferSize: anInteger codec: aGRCodec
+	^ self basicNew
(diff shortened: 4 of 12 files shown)


theseion:
@jbrichau I have `WAComboResponse` working with streaming in Pharo 9 and 10, but below the converters don't behave like I need them too (i.e., there is no converter that can write strings to a `ByteArray`. In Pharo 10 Zinc does that and in Pharo 9 the UTF-8 converter does it too). I don't know how I should solve it. Maybe we need another delegating stream that converts from characters to bytes but I'm reluctant to do that.

repository/Seaside-Tests-Core.package/WAComboResponseTest.class/instance/response.st
```diff
@@ -1,3 +1,7 @@
 accessing
 response
-	^ super response ifNil: [ response := WAComboResponse external: (GRPlatform current writeCharacterStreamOn: (String new: 4096)) ]
\ No newline at end of file
+	^ super response ifNil: [
+		response := WAComboResponse
+			onStream: (GRPlatform current writeCharacterStreamOn: (ByteArray new: 1024))
```
marschall, line 5: This doesn't look right. `#writeCharacterStreamOn:` wants a `String`, should we instead send `#writeCharacterStreamOn:`?

repository/Seaside-Core.package/WAComboResponse.class/instance/close.st
```diff
@@ -3,5 +3,14 @@ close
 	"Send a zero-sized chunk to end the data transfer."
 
 	closed ifTrue: [ self error: 'Response is closed' ].
-	externalStream nextPutAll: '0'; crlf; crlf; flush.
+	externalStream
+		"zero length chunk"
+		nextPut: 0;
+		"crlf"
+		nextPut: 13;
```
marschall, line 10: I don't know how portable this is but one could also consider

    Character cr greaseInteger

repository/Seaside-Core.package/WAComboResponse.class/instance/commit.st
```diff
@@ -8,7 +9,20 @@ commit
...
-	self writeStatusOn: externalStream.
-	self writeHeadersOn: externalStream.
-	self writeCookiesOn: externalStream.
-	externalStream crlf; flush
\ No newline at end of file
+	
+	"The external stream is usually byte based. Unfortunately, the
+	streams aren't compatible enough at the moment such that everything
+	could be written to the external stream directly (in particular, writing of
+	singel characters doesn't work). Luckily, the encoders already translate
```
marschall, line 16: single

marschall:
> (i.e., there is no converter that can write strings to a ByteArray. In Pharo 10 Zinc does that and in Pharo 9 the UTF-8 converter does it too). I don't know how I should solve it. Maybe we need another delegating stream that converts from characters to bytes but I'm reluctant to do that.

I think for 4.0 we should bite the bullet and make the codecs `String` -> `ByteArray`

marschall:
LGTM apart from my comments.

repository/Seaside-Tests-Core.package/WAComboResponseTest.class/instance/response.st
theseion: It should be something else, yes. I presume, you meant to write `writeBinaryStreamOn:`?

repository/Seaside-Core.package/WAComboResponse.class/instance/close.st
theseion: Yes. I wanted this to be fast though.

marschall: Fair enough

repository/Seaside-Tests-Core.package/WAComboResponseTest.class/instance/response.st
marschall: > I presume, you meant to write writeBinaryStreamOn:?

Yes

> I remember having to fight a lot with the wrapping streams (i.e., Zinc).

Let's figure out the problem and solve it.

theseion: `#writeCharacterStreamOn:` simply delegates to `WriteStream on:`, which operates on collections, so in that sense it's fine. Also, this is in a test and there is no `#writeBinaryStreamOn:` so far. I'll just leave it if you don't object.

PR #1249, opened 2021-03-30
port Seaside to 3.7.0

### LINKED ISSUE #1248: port Seaside to 3.7.0
initial Seaside test runs against 3.7.0 have quite a few failures:
```
**************************************************************************************
	Results for Seaside3 tests
1447 run, 1304 passes, 0 expected defects, 128 failures, 15 errors, 0 unexpected passes
**************************************************************************************
```
failures likely due to a change to one of the fundamental GemStone methods ... the handling of breakpoints has changed, which was the source of GLASS failures, but the GLASS tests are running clean at this point ...

dalehenrich:
fix #1248 ... just needed an update to the baseline ..


jbrichau (approved):

jbrichau:
Seems like 3.7.0 has consistent test failures. Not sure if you want me to merge it in right away or you want to hit the button yourself. I'll wait a little ;-)

dalehenrich:
It was the 3.6.0 test that I wanted to rerun when I had simply created a new branch based off of master with no changes at all and the only change I've made here is to add 3.7, 3.8, and 3.9 to the baseline ... so something has changed on servers(?) to result in this consistent failure...

Not sure why the one test is failing for 3.6.0, since it doesn't fail locally ... the line where it is failing appears to be the statusGems report and that particular assertion is used in a number of of different GemServer tests that are run in the same test suite ... so it is a mystery compounded by the fact that in order to debug I cannot run just the 3.6.0 test ... and it doesn't seem that i have a private copy of the .yml test driver (at least I couldn't find a copy that I could edit) ... I'm guessing that it has to do with server loading ... but that's just one guess among many ... any advice on a way to proceed?

jbrichau:
@dalehenrich seems like it did run in your fork: https://github.com/dalehenrich/Seaside/actions/runs/699767589

I'm guessing it's a spurious error and the chances were not on the right side when the tests ran. I'll retrigger this batch again.

dalehenrich:
I've got some suspicions about the restartGem code and it's behavior on a heavily loaded machine ... I don't think that the stopGem code removes the pid file and if a gem was slow enough in getting started, the test could be looking for a process that has just died ... at least that's my suspicion ... thanks for merging !

dalehenrich:
regarding the restartGem, the stop code DOES wait for the pid to go away and then it deletes the file ... so somethings else is afoot ...

PR #1252, opened 2021-06-28
Fixes WARenderContext using the wrong visitor for WAHtmlCanvas

eMaringolo:
Rendering components directly from from a `WABuilder` skipped visiting the decorations (if any) of the rendered components.

So in the following code:
```smalltalk
  | component |
  component := (SomeComponent new) 
                             addDecoration: (SomeDecoration new);
                             yourself.

 WAHtmlCanvas builder
     fullDocument: true;
     rootBlock: [ :root | component updateRoot: root ];
     render: [ :html | html render: component ]
```

The instance of `SomeComponent` will be rendered without its decorations.

The issue comes from the `WARenderContext` being initialized generically in `WABuilder` but with a specific visitor (`WARenderingGuide`) in `WARenderPhaseContinuation`.

I propose a change that is idempotent to other kind of canvases (XML, JSON, etc.) , but gives a method in `WAHtmlCanvas` to hook in the initialization of the render context to set the proper visitor.

I include a test case: `WABuilderCanvasTest>>#testBodyWithDecorations`

### DIFF
--- a/repository/Seaside-Canvas.package/WAHtmlCanvas.class/class/configureRenderContext..st
+++ b/repository/Seaside-Canvas.package/WAHtmlCanvas.class/class/configureRenderContext..st
@@ -1,0 +1,5 @@
+delegation
+configureRenderContext: aWARenderContext
+	"We set aWARenderContext to use a rendering guide as visitor."
+		
+	aWARenderContext visitor: (WARenderingGuide client: aWARenderContext visitor).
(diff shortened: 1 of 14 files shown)


eMaringolo:
What's wrong with these platforms failing? Why does that happen?

jbrichau:
@eMaringolo Squeak is an accepted failure but the GemStone 3.6.1 is showing a test that is failing regularly and we should fix it. It does not have anything to do with this PR though.
Thanks for sending this in. I'll have a spin with it asap.

PR #1272, opened 2021-08-02
Adds support for HEAD and PATCH route definition using method pragmas

### LINKED ISSUE #1271: Add HEAD and PATCH as supported HTTP methods in WARouteBuilder
`WARouteBuilder` there is no support for `HEAD` or `PATCH`  ​method handling.

eMaringolo:
Fixes #1271


eMaringolo:
I think this should be on-hold. It seems that implementing HEAD in the handler is not completely correct. The  [HTTP spec](https://httpwg.org/specs/rfc7231.html#HEAD) says: "The server _SHOULD_ send the same header fields in response to a HEAD request as it would have sent if the request had been a GET". 

With this, if there is a HEAD request and no matching route is defined for HEAD method (but there are matching routes for GET), then it will return a 404, when it should fall back to processing the GET matching route but discarding it's body.

eMaringolo:
I added that retry to my latest commit in this PR. I don't know whether such retry should also be present in `WARestfulFilter`, I think it shouldn't, since you cannot perform a HEAD that is idempotent to an URL with an action callback id.

jbrichau:
Simply retrying the `HEAD` request as a `GET` request, will execute the request and return a body and it is not supposed to do that. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

Although not explicitly forbidden, it does mean that this implementation is not useful to check the headers of a request without executing the actual request. 

>For example, if a URL might produce a large download, a HEAD request could read its Content-Length header to check the filesize without actually downloading the file.

jbrichau:
Digging a little further than the mozilla developer pages: according to the RFC, it is actually forbidden:

>The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

eMaringolo:
I think there is a misunderstanding here, I'm not saying that we retry the HEAD as a GET and return the full GET response, but instead to retry the HEAD as a GET and then trim the GET body response, effectively returning only the HEAD.

In any case this is not priority, and was more as a side request of a user trying to handle HEAD directly in a `WARestfulHandler` and not being able to do so, because HEAD requests cannot be handled by Seaside-REST, and instead rely on the server adaptor.

jbrichau:
The change to `WARestfulHandler` is trying the `GET` request in case no route was found for the `HEAD` request, no? Am I reading it that wrong? I have not loaded the code in an image yet to play.

eMaringolo:
You're correct, that's what it is doing. I found no other way to do it, but it feels too hacky.

In the case of Zinc, you can see that in `handleRequest:timing:` it performs a regular request but then clears the entity of the response.

However, the more we talk about this, the more I realize this might be wrong from the beginning, if there is no GET route defined, then there shouldn't be any HEAD route handling. And so, HEAD handling is something that belongs more to the server (Zinc, SST, etc.) than to any Seaside adaptor or request handler.

I think we should close this PR.

theseion:
While the HTTP RFC says that `GET` retrieves information and `HEAD` is the same as `GET` without body, we all know that many `GET` requests will actually modify state. With this in mind we can easily construct an example where a `HEAD` request will modify the state and the client may miss out on the return data. I can see this as an argument to at least disable the `HEAD` method for a given `GET` route (opposed to completely disabling `HEAD`).

I agree that we don't need to have a handler for `HEAD` if there is no equivalent handler for `GET`. However, I think there is a case to be made for defining a `HEAD` handler for an accompanying `GET` handler. This would make it possible to, for example, prevent expensive computations for return data.
Of course, then you have to deal with all of the edge cases, e.g. what to do when there's a `HEAD` handler but nod `GET` handler etc.

eMaringolo:
If we stick to the HTTP spec, then `GET` should be idempotent and do not modify state. To avoid that, if you try to run a `HEAD` request through a `WAApplication` then it will return a `501 (Not implemented)` status response, because running such request could trigger action callbacks.

So the only use case here (that kickstarted this feature request) would be the one you just described, somebody wanting to serve a `HEAD` request without having the server running an expensive computation, so a slightly variation of the GET (but with a `Content-Length: 0` response header, which might not be accurate, but is not forbidden by the spec.

So, given that there is no real need, and any solution is not elegant, I'm closing this PR. I'll likely re-submit one to add `PATCH` method support.

jbrichau:
@eMaringolo I think we can could have left in the support, just not the automated conversion of `HEAD` requests to `GET` requests.

eMaringolo:
This continues to get back to me, and I would like to settle on something useful.

I realized that without that "retry hack" a `HEAD` request will never find a matching route (defined as `GET` method), so any `HEAD` request will return a `404` (you can try it locally).

What would be a good alternative to this retry hack?
To make `GET` routes also match `HEAD` requests? 

In `WAApplication` it is explicitly handled, returning a `501 (Not Implemented)` but in other handlers is wrongly returning a `404` when there is a matching `GET` route.

Issue #1270, opened 2021-08-02
Implement CORS request filter


eMaringolo:
I'm creating this issue as a placeholder, since I have a `WACORSFilter` that I plan to submit back to this repository soon.

I'm polishing it to make it more user friendly (beyond my basic needs) and I added a functional test to make it work (it involves setting up an adapter on another port to actually need CORS handling.

As such, the filter is meant to be used on any request handler (app, session, rest), although it would certainly make more sense in REST contexts.

jbrichau:
This one got merged last year :-)

PR #1274, opened 2021-08-11
Adds CORS filter and Functional test

eMaringolo:
I implemented a basic CORS filter with a functional test (including setting up a separate server adaptor to handle cross-origin requests).

It would be great to merge this into the upcoming release, so it can be ported to VAST (and other dialects).

### DIFF
--- a/repository/Seaside-Core.package/WACORSFilter.class/README.md
+++ b/repository/Seaside-Core.package/WACORSFilter.class/README.md
@@ -1,0 +1,1 @@
+Implements a WARequestFilter that adds support to handle CORS requests.
--- a/repository/Seaside-Core.package/WACORSFilter.class/instance/addAllowedMethod..st
+++ b/repository/Seaside-Core.package/WACORSFilter.class/instance/addAllowedMethod..st
@@ -1,0 +1,6 @@
+methods
+addAllowedMethod: httpMethod
+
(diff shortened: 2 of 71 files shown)


eMaringolo:
This solves #1270

eMaringolo:
@jbrichau Can this be merged? :-)

jbrichau:
> @jbrichau Can this be merged? :-)

Sorry for the unresponsiveness. I'm swimming in too much things at once. I will test this hopefully this week!

jbrichau:
@eMaringolo I did not notice before that the Gemstone builds are failing here. This is probably because of incompatibility. We will need to fix this before this can be merged.

eMaringolo:
@jbrichau But the incompatibility is caused by something else, right? Looking at the job details, the failing tests have nothing to do with my proposed changes, or do they?

jbrichau:
The failing tests are `testSentButNotImplemented` and `testUndefinedSymbols`. These tests fail because there is code added that sends unimplemented messages and references undefined symbols. 

I'm guessing it's in the example code, which is not part of a test so no other tests will fail.

`expandMacrosWith:` is probably the unimplemented message in GemStone.

jbrichau:
@eMaringolo I merged this in a branch to be able to debug and fix the broken tests in GemStone.

Issue #1291, opened 2021-11-19
WAMutexTest>>testSelfTerminate ... what behavior is being tested?


dalehenrich:
This test as written:
```
testSelfTerminate
	| value semaphore |
	value := nil.
	semaphore := GRPlatform current semaphoreClass new.
	process := [ 
	value := mutex
		critical: [ 
			semaphore signal.
			mutex terminateOwner.
			1 ] ]
		newProcess.
	process resume.
	semaphore wait.
	self assert: mutex owner isNil.
	self assert: value isNil.
	self assert: (GRPlatform current isProcessTerminated: process)
```
presumably has been passing in Pharo and GemStone for a long time ... However, in 3.7.0 (upcoming release) we have done a fair amount of work to improve the process scheduling and termination model ... specifically targeting the handling of ensure blocks ... we weren't quite up to snuff for all cases.

We had to add a Delay after the `semaphore` wait in order for the "test to pass":
```
testSelfTerminate
	| value semaphore |
	value := nil.
	semaphore := GRPlatform current semaphoreClass new.
	process := [ 
	value := mutex
		critical: [ 
			semaphore signal.
			mutex terminateOwner.
			1 ] ]
		newProcess.
	process resume.
	semaphore wait.
"-->"	[ process isTerminated ] whileFalse: [ Delay waitForMilliseconds: 20 ].
	self assert: mutex owner isNil.
	self assert: value isNil.
	self assert: (GRPlatform current isProcessTerminated: process)
```
In our new and improved implementation, the process waiting on the semaphore is allowed to run as soon as the `semaphore signal` is sent ... and as a result the mutex owner isn't cleared by the time the assertion is made and this is an explicit change in behavior we made, because we think it is the correct semantics for a process to be allowed to start running as soon as the semaphore is signaled, rather than wait it's turn in the scheduler queue and wait for the running process to be terminated before allowing another process to run ...

From the test it isn't clear whether it is important:
1. that the mutex owner is cleared before any other process gets a chance to run, OR
2. that the mutex owner is (eventually) cleared upon process termination

If order of process execution is important then it will be necessary for us to not only change this test for GemStone, but to make changes in the other parts of the system where this behavior is important ...

As you know, there is probably not a whole lot of multi-process operations that should be done in a GemStone Seaside implementation, but I don't want to just change this test and cross our fingers :)

dalehenrich:
There are actually 2 tests that are failing since our changes to GsProcess:
```
[...]

PR #1296, opened 2021-12-26
Fix WAMutexTest>>testSelfTerminate

jbrichau:
Use process priority rather than semaphore to ensure the spawned process is terminated in the test case.
See #1291 for more info.
Thanks @martinmcclure for the test fix!


jbrichau:
Merging this one in (I was holding that off until fixing the broken builds for GemStone 3.6, but turns out that is going to take a little longer in https://github.com/SeasideSt/Seaside/issues/1297)

PR #1308, opened 2022-04-09
Fix#1307 - Replace obsolete message send with functional equivalent

johnhwoods:
WAPharoWalkback resulted in "Internal Error" on Pharo 10 when  'renderContentOn:' is incorrect.

This is because the variable listing method referred to Context>>tempNamedAt: which was deprecated in 9.0 and has gone in 10.0.  Replaced with functionally equivalent iteration over temp vars.


jbrichau:
Thx @johnhwoods for sending this in. I had a slightly different stab at in https://github.com/SeasideSt/Seaside/commit/722c2cc9cc4973e29f18ccdf6dcf2f666b0bd2ca 
Closing this PR.

PR #1322, opened 2022-08-16
Websockets

jbrichau:
Added support for Websockets in Seaside.


repository/Seaside-WebSocket-Core.package/JSWebSocket.class/instance/javascriptContentOn..st
```diff
@@ -0,0 +1,8 @@
+printing
+javascriptContentOn: aStream
+	"TODO: window.location should not be retrieved here..."
```
marschall, line 3: Should we use `#baseUrl` or something similar here instead?

repository/Seaside-WebSocket-Core.package/WARequest.extension/instance/isWebSocketSetupRequest.st
```diff
@@ -0,0 +1,9 @@
+*Seaside-WebSocket-Core
+isWebSocketSetupRequest
+	"Return true when request can be considered a valid WebSocket setup request"
+	
+	^ self method = #GET
```
marschall, line 5: Should probably use `#isGet`

repository/Seaside-WebSocket-Core.package/WARequest.extension/instance/isWebSocketSetupRequest.st
```diff
@@ -0,0 +1,9 @@
...
+			and: [ (self headers at: 'upgrade' ifAbsent: [ ^ false ]) asLowercase = 'websocket'
```
marschall, line 6: `#headerAt:ifAbsent:` may be a bit shorter

repository/Seaside-WebSocket-Core.package/WAWebSocket.class/README.md
```diff
@@ -0,0 +1 @@
+I am a WebSocket.
```
marschall, line 1: More an abstract class with server specific implementations.

repository/Seaside-WebSocket-Core.package/WAWebSocketJSPusher.class/instance/websocketEventHandlerJavaScriptOn..st
```diff
@@ -0,0 +1,4 @@
+protected
+websocketEventHandlerJavaScriptOn: aStream
+	aStream 
+		nextPutAll: '.addEventListener(''message'', (event) => { Function(`"use strict";return (${event.data})`)(); })'
```
marschall, line 4: why the need for a function?

repository/Seaside-WebSocket-Core.package/WAWebSocketPusher.class/instance/sockets.st
```diff
@@ -0,0 +1,5 @@
+accessing
+sockets
+	"Answer the current list of sockets."
+
+	^ sockets
```
marschall, line 5: Isn't this error prone because the caller has to care about the mutex? Shouldn't we just offer `#socketsDo:`?

repository/Seaside-WebSocket-Core.package/WAWebSocketPusher.class/instance/socketsDo..st
```diff
@@ -0,0 +1,4 @@
+protected
+socketsDo: aOneArgumentBlock
+	self mutex critical: [
```
marschall, line 3: Would a `^` be helpful here so the sender can easily access the return value from the provided block?

repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/responseFrom..st
```diff
@@ -1,6 +1,9 @@
 converting
 responseFrom: aRequestContext
 	| partialHeaders cookies fullHeaders seasideResponse contents entity contentType |
+	"TODO: is websocket only... useful/needed to package in separate package?"
+	(aRequestContext properties at: #webSocket ifAbsent:[ nil ])
```
marschall, line 5: I noticed the formatting in a lot of places omits a space after `:`. Is this intentional?

jbrichau: I have a lot to clean up still and there’s a lot of copy paste for now. I made a draft PR to follow up on the CI build. So most of this code is not in the final clean state, rather fast prototyping.

repository/Seaside-WebSocket-Examples.package/WAWebSocketExampleBroadcaster.class/instance/handleWebSocketSetup.in..st
```diff
@@ -0,0 +1,19 @@
+handling
+handleWebSocketSetup: aWAWebSocket in: aRequestContext
+	
+	"webSocket timeout: 300." "5 minutes timeout"
+	
+	Transcript cr; show: 'New ', aWAWebSocket printString.
+
+	aWAWebSocket onMessage:[ :data |
+		self socketsDo: [ :socket | socket send: data ] ].
```
marschall, line 9: I'm a bit concerned about the socket iteration pattern.

- In this case, and probably others, since `#socketsDo:` acquires a mutex it blocks until data has been sent to all sockets. I wonder if it would be beneficial to have some sort of job/task queue and submitting a task for every socket and hold the mutex only for submitting the tasks.
- Between when we decided to send data to the sockets and actually doing it the socket may have become closed or invalid which makes we wonder whether we should:
  - check each web socket for validity in `#send:` and possibly removing it
  - wrap each `#send:` in an `#on:do:`
- Having all the sockets in an array, linearly scanning over them and hoping `#onClose:` will always get sent and the web socket removed. I'm concerned with scalability and that we end up with dead or closed web sockets in the array that never get removed.

repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/responseFrom..st
marschall, line 5: Should we do something special when shutting down the server?

repository/Seaside-WebSocket-Core.package/WAWebSocketPusher.class/instance/addSocket..st
```diff
@@ -0,0 +1,3 @@
+accessing
+addSocket: aWebSocket
+	mutex critical: [ sockets := sockets copyWith: aWebSocket ]
```
marschall, line 3: Is the intention of using CoW here that readonly access to the sockets (eg. just iterating) can be done without a mutex?

repository/Seaside-WebSocket-Core.package/WAWebSocketJSPusher.class/instance/websocketEventHandlerJavaScriptOn..st
jbrichau: Because 'eval' is bad :-). https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!

repository/Seaside-WebSocket-Core.package/JSWebSocket.class/instance/javascriptContentOn..st
jbrichau: Indeed. It should probably use the settings of WARequestHandlingConfiguration... it's just that I do not prefer that way of working. Alas, I see no alternative either.

repository/Seaside-WebSocket-Examples.package/WAWebSocketExampleBroadcaster.class/instance/handleWebSocketSetup.in..st
jbrichau: yes, all valid concerns and this still needs to be addressed. In addition, the seaside session needs to be kept alive as long as the sockets are communicating/existing. While doing experiments with the simple counter example, I encounter websocket connections being closed without any callback firing without a consistent timeout or other immediately visible reason. So, I still have a long way to go to make sure this works in practice.

PR #1327, opened 2022-08-21
Remove old FileLirary methods

marschall:
Remove old FileLirary methods used for migrations long ago.


marschall:
Strictly speaking these methods aren't deprecated at the moment.

They were used either for migrating to Seaside 2.9 or get rid of the literal caching that we have moved away from a long time ago.

jbrichau (approved):

PR #1334, opened 2022-08-23
Fix WAVersionUploader

marschall:
Fixes the version uploader on Pharo 10.

- `RWBinaryOrTextStream` was undeclared
- `#newVersion` was not implemented

### DIFF
--- a/repository/Seaside-Pharo-Tools-Web.package/WAVersionUploader.class/instance/newVersion..st
+++ b/repository/Seaside-Pharo-Tools-Web.package/WAVersionUploader.class/instance/newVersion..st
@@ -1,10 +1,10 @@
 actions
 newVersion: aWorkingCopy
 	| version stream |
-	version := [ aWorkingCopy newVersion ]
+	version := [ aWorkingCopy newVersionIn: aWorkingCopy repositoryGroup ]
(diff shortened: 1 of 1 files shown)


repository/Seaside-Pharo-Tools-Web.package/WAVersionUploader.class/instance/newVersion..st
```diff
@@ -1,10 +1,10 @@
 actions
 newVersion: aWorkingCopy
 	| version stream |
-	version := [ aWorkingCopy newVersion ]
+	version := [ aWorkingCopy newVersionIn: aWorkingCopy repositoryGroup ]
```
marschall, line 4: I'm not sure whether or `#repositoryGroup` or `#theCachedRepository` or something else is correct.

jbrichau: I have not investigated further, but seems like this is also how it's done in MetacelloPharoPlatform and the MCWorkingCopyTest. So looks fine to me.

I have never really understood the use of the tool though...

jbrichau (approved):

Issue #1333, opened 2022-08-23
WAScreenshot is broken


marschall:
WAScreenshot is broken in Pharo. At least with Pharo 10 on macOs:

- the screenshot is empty
- `WAScreenshot >> #handleEvent:` references `Sensor` which is undeclared, it used to be an instance of `EventSenor`
- `WAScreenshot >> #handleEvent:` sends `#handleEvent:` or `#processEvent:`, neither of which seem implement or working with the array created in `WAScreenshot >> click:`

jbrichau:
Was fixed in https://github.com/SeasideSt/Seaside/pull/1335

PR #1335, opened 2022-08-23
Fix displaying screenshot of Pharo

marschall:


marschall:
This only fixes the displaying of a screenshot (see #1333), the clicks are still broken.

repository/Seaside-Pharo-Tools-Web.package/WAScreenshot.class/instance/form.st
```diff
@@ -1,5 +1,5 @@
 accessing
 form
 	^ zoom
-		ifFalse: [ Display copy ]
+		ifFalse: [ UIManager default world imageForm ]
```
marschall, line 4: This was more or less trial and error. I'm not sure this is actually the right way.

jbrichau: If it works, it works

jbrichau (approved):

PR #1344, opened 2022-09-13
Modifies WAServerManager to support subclasses to be defined as default

eMaringolo:
VAST, and probably other dialects, have a subclass of WAServerManager that is used instead of it.

Since `WAServerManager[default]` is a class instance variable, then sending the message `default` to `WAServerManager` will always answer an instance of itself. Which is more troubling if you access it by means of `WAAdmin defaultServerManager`.

This PR modifies the `default` class instance variable to be a class variable (named `Default`), and provides a `beDefault` method that can be used on subclasses to be the "singleton" default.

### DIFF
--- a/repository/Seaside-Core.package/WAServerManager.class/class/default.st
+++ b/repository/Seaside-Core.package/WAServerManager.class/class/default.st
@@ -1,4 +1,6 @@
 accessing
 default
-	default contents ifNil: [ default contents: self new ].
-	^ default contents
+
+	self initializeDefaultValueHolder.
+	Default contents ifNil: [ Default contents: self new ].
+	^ Default contents
--- a/repository/Seaside-Core.package/WAServerManager.class/class/beDefault.st
+++ b/repository/Seaside-Core.package/WAServerManager.class/class/beDefault.st
@@ -1,0 +1,5 @@
+configuring
+beDefault
+	"Set an instance of receiver to be the default Server Manager."
+	
+	Default contents: self new
--- a/repository/Seaside-Core.package/WAServerManager.class/class/beDefault.st
+++ b/repository/Seaside-Core.package/WAServerManager.class/class/beDefault.st
@@ -2,4 +2,4 @@ configuring
 beDefault
 	"Set an instance of receiver to be the default Server Manager."
 	
-	Default contents: self new
+	self setDefault: self new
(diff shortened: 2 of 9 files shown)


marschall:
Out of curiosity, why do you need to subclass `WAServerManager`? Is there something we can do so that you don't have to do this?

eMaringolo:
The `WASstServerManager` in VAST provides logging of the start/stop events and also a callback interface (think of an "Announcer") to notify these events to the visual control panel can reflect the state without unnecessary polling.

repository/Seaside-Core.package/WAServerManager.class/class/default.st
```diff
@@ -1,4 +1,6 @@
 accessing
 default
-	default contents ifNil: [ default contents: self new ].
-	^ default contents
\ No newline at end of file
+
+	self initializeDefaultValueHolder.
```
jbrichau, line 4: Perhaps renaming the method `initializeDefaultValueHolder` to `ensureDefaultValueHolder` makes it clear that `Default` is not reinitialized when invoking this method? 
I was specifically triggered by seeing this method call here while the value of `Default` should remain the same `WAValueHolder` instance (or `TransientValue` in GemStone)

jbrichau (changes requested):
Hi Esteban. Thanks for this and sorry for having skipped this PR several weeks ago.
I just wanted to ask a name change for the method `initializeDefaultValueHolder` to reflect it does not reinitialize. I suggest `ensureDefaultValueHolder`.

I'm all pro using class vars rather than class instvars for this. I'm not sure if this was a mistake or deliberate choice. It seems strange for a 'default' singleton implemention though, so I suspect the former.

repository/Seaside-Core.package/WAServerManager.class/class/default.st
jbrichau: Hey @eMaringolo. What do you think about my suggestion?

eMaringolo: Sorry, I missed this. I'm okay with your suggestion, I think it is more intention revealing that the one with the `initialize` prefix.

jbrichau: I will merge and rename

Issue #1376, opened 2023-09-06
WAFastCGIRequestConverter>>#requestUrlFor: signals a WAInvalidUrlSyntaxError when the ‘REQUEST_URI’ is ‘/?://’


Rinzwind:
When the ‘REQUEST_URI’ parameter is `/?://` in [`WAFastCGIRequestConverter>>#requestUrlFor:`](https://github.com/SeasideSt/Seaside/blob/7d0ccbd436f4e7489f3ae847dcfe398ee5a2181d/repository/Seaside-Adaptors-FastCGI.package/WAFastCGIRequestConverter.class/instance/requestUrlFor..st), a WAInvalidUrlSyntaxError is signaled, although `/?://` is a valid ‘origin-form’ per [section 3.2.1 in RFC 9112](https://www.rfc-editor.org/rfc/rfc9112.html#name-origin-form).

The relevant rules from [RFC 9112](https://www.rfc-editor.org/rfc/rfc9112.html#appendix-A), [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#appendix-A) and [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#appendix-A) are:

```abnf
origin-form   = absolute-path [ "?" query ]

absolute-path = 1*( "/" segment )

segment       = *pchar

query         = *( pchar / "/" / "?" )

pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

PR #1377, opened 2023-09-08
Fix #initializeFromString: in WAUrl to take into account that ‘://’ can also be included in the query

Rinzwind:
This pull request fixes issue #1376 by making `#initializeFromString:` in WAUrl take into account that `://` can also be included in the query.

The example this pull request adds to `#testNastyEdgeCasesParsing` reflects the use of `#absolute:` in `#requestUrlFor:` in WAFastCGIRequestConverter: the argument to `#absolute:` is not a valid absolute URI as defined in [section 4.3 in RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3), as it does not begin with a scheme, but it is a valid request target as defined in [section 3.2 in RFC 9112](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2). WAUrl and WAFastCGIRequestConverter should maybe be made to better distinguish between those.


jbrichau (approved):
Looks good to me.

PR #1380, opened 2023-09-27
Re-introduce escaping of a closing tag when encoding javascript

### LINKED ISSUE #1379: Nested script tags are not properly escaped
Adding the following snippet to a Seaside render method breaks the generated html because the closing script tag in the jQuery append expression is not properly escaped.

```
html anchor script:
	((html jQuery this closest: 'div') append: [ :r |
		 r div script: (html jQuery this
				  on: 'click'
				  selector: '.class'
				  do: (JSStream on: 'alert(''nested script''')) ])
```
Screenshot 2023-09-27 at 17 35 08

In Seaside 3.0, the method https://github.com/SeasideSt/Seaside/blob/61f25aa0e8b820cf1e3d554ef8bcceb12e307233/Javascript-Core.package/JSStream.class/class/encodeString.on..st used to contain code that escapes closing of nested tags:

```Smalltalk
"avoid that browsers mistakenly take the output as a closing tag"
(last = $< and: [ char = $/ ])
	ifTrue: [ aStream nextPutAll: '\/' ]  
	ifFalse: [ aStream nextPut: char ] ]
```

https://github.com/seasidest/seaside/issues/726, the commit comment that removed this from Seaside 3.0 code mentions:

> do we really have to encode `</` as `<\/`?
>  * if it's inside a <script> it should work fine without it
>  * if it's inside anything other and <script> or <style> HTML escaping should kick in
>  * if it's inside an attribute value HTML escaping should kick in

jbrichau:
Fixes #1379


Rinzwind:
There’s a note in [section ‘4.12.1.3 Restrictions for contents of script elements’ in the HTML Living Standard](https://html.spec.whatwg.org/commit-snapshots/a5334644b892d4406f44869c3f738ad68ecd44a7/#restrictions-for-contents-of-script-elements) which recommends escaping `<!--` and `<script` as well. The following snippet is based on the example in the section, the page does not show ‘Test Paragraph’ as would be expected:

```smalltalk
FileLocator imageDirectory / 'Test.html' writeStreamDo: [ :stream |
   WAHtmlCanvas builder fullDocument: true;
      render: [ :html |
         html
            script: (html logger log: '<!-- <script>');
            paragraph: 'Test Paragraph' ]
      on: stream ]

```

jbrichau:
@Rinzwind Good point. I moved your comment to the issue tracker since it requires a bit more work than to re-introduce the default escaping of all closing tags. To fix the issue, that should suffice but I consider it a temporary fix while we work on following that advice in the link.

PR #1384, opened 2023-10-29
fix: Add PackageManifest stub class

### LINKED ISSUE #1382: PackageManifest extensions should not be loaded in Squeak
Loading Seaside in Squeak trunk causes warnings due to extensions on `PackageManifest` classes that can't be loaded because the classes don't exist there.

theseion:
Add PackageManifest to prevent warnings when loading Pharo package manifest subclasses.

Fixes #1382


jbrichau:
@theseion The package should also be added to the metacello baseline for Squeak

theseion:
It's already there, isn't it?

```smalltalk
baselinesqueak: spec

	"Mind that baselinepharo still has a squeakcommon section!"
	spec
		for: #squeak
		do: [ 
			spec
				package: 'Seaside-Squeak-Compatibility';
        "..."
```

jbrichau:
@theseion Ow. I misread this as a new package but indeed. It already is there.

jbrichau (approved):

Issue #1213, opened 2020-07-11
Zinc-Seaside adaptor has no unit tests


jbrichau:
To prevent more bugs (like #1211), we need some unit tests to verify the implementation of the Zinc-Seaside adaptor.

PR #1387, opened 2023-11-02
Fix #requestHeadersFor: on ZnZincServerAdaptor to combine field values for repeated field names

Rinzwind:
This pull request fixes `#requestHeadersFor:` on ZnZincServerAdaptor to combine field values for repeated header field names.

[Section ‘5.2. Field Lines and Combined Field Value’ in RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.2) states:

> When a field name is only present once in a section, the combined "field value" for that field consists of the corresponding field line value. When a field name is repeated within a section, its combined field value consists of the list of corresponding field line values within that section, concatenated in order, with each field line value separated by a comma.

See also [section ‘6.3. Header Fields’](https://www.rfc-editor.org/rfc/rfc9110.html#section-6.3).

Example:

```smalltalk
request := String crlf join: #('GET / HTTP/1.1'
	'Accept: application/xml'
	'Accept: application/json'
	'' '').
(ZnZincServerAdaptor new requestFor: (ZnRequest readFromString: request))
	headerAt: 'accept'
```

Without this fix, the example returns `'application/json'`.
With this fix, the example returns `'application/xml,application/json'`.

### DIFF
--- a/repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/requestHeadersFor..st
+++ b/repository/Zinc-Seaside.package/ZnZincServerAdaptor.class/instance/requestHeadersFor..st
@@ -3,5 +3,10 @@ requestHeadersFor: aZincRequest
 	| fields |
 	fields := Dictionary new.
 	aZincRequest headersDo: [ :key :value |
-		fields at: key asLowercase put: value ].
+		| keyLowercase combinedValue |
+		keyLowercase := key asLowercase.
+		combinedValue := fields at: keyLowercase
+			ifPresent: [ :presentValue | presentValue , ',' , value ]
+			ifAbsent: [ value ].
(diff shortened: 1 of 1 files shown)


theseion (commented):
LGTM

jbrichau:
This reminds me of https://github.com/SeasideSt/Seaside/issues/1211 which the exact same issue for responses... It does show how we need unit tests for these adaptors (https://github.com/SeasideSt/Seaside/issues/1213)

Rinzwind:
Reading issue #1211 makes me wonder: shouldn’t ZnZincServerAdaptor’s `#requestHeadersFor:` return a WAHeaderFields instead of a Dictionary? Then the separate headers can be preserved. The methods in WARequest’s ‘accessing-headers’ protocol that answer a collection should then combine them.

PR #1402, opened 2024-02-07
Update UTF-8 sampler link

eMaringolo:
The UTF-8 sampler moved to a new location, updates the functional test to point there


jbrichau:
Thanks!

PR #1419, opened 2024-07-06
Add support for web components

### LINKED ISSUE #1029: Add support for <template> and <slot>
We should add support for the <template> and <slot> elements as well as the slot attribute either in the Seaside-Canvas package or a new Seaside-WebComponents package

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot

marschall:
This PR includes two changes:

1. Add support for <template> and <slot>
2. Add support for rendering applications without <html>, <head> and <body> elements. The use case for this is if you want to implement a web component that ends up calling a Seaside application.

Documentation: https://github.com/SeasideSt/Seaside/wiki/WebComponents

Fixes #1029


jbrichau:
Interesting. I have not dabbled with Web Components yet... could you include an example on how this would be used?

repository/Seaside-Canvas.package/WASlotTag.class/README.md
```diff
@@ -0,0 +1 @@
+The slot HTML element—part of the Web Components technology suite—is a placeholder inside a web component that you can fill with your own markup, which lets you create separate DOM trees and present them together.
```
jbrichau, line 1: Wouldn't we also need to add the `slot` attribute to the `WATagBrush` ?

marschall:
> Interesting. I have not dabbled with Web Components yet... could you include an example on how this would be used?

Embedding Seaside components in a non-Seaside page without resorting to frames.

```html
<!DOCTYPE html>
<html>
  <head>
    <!-- -->
    <script src="seaside-components.js" defer></script>
  </head>
  <body>
    <!-- -->
    <seaside-component url="/examples/counter">Loading...</seaside-component>
    <!-- -->
  </body>
</html>
```

The components could be ajaxified so they could make use of full page refreshes without reloading the whole page.

I'll work on a demo.

jbrichau:
Oh ok, I think I get it. The 'seaside-component.js' implementation uses the server-side to render the html contents? 

ps: see you tomorrow ;-)

marschall:
> The 'seaside-component.js' implementation uses the server-side to render the html contents?

Yeah, exactly. I added some demo code in an additional commit. Still not sture what the ideal way to include custom scripts and styles is.

adriaon:
💜

PR #1421, opened 2024-07-06
Fix browser deprecations in Pharo 11

marschall:
This should help with all the deprecation warnings on Pharo 12 in logs.


jbrichau (approved):
thanks!

PR #1434, opened 2024-07-13
Fix html deprecations

### LINKED ISSUE #1418: Deprecated messages in functional tests
When running the functional tests several deprecated messages are sent:

- `WARenderLoopContinuation>>#updateRoot:` sends `#charset:` to `WAMetaElement`
- `WACanvasTableFunctionalTest` sends
  - `WATableDateTag>>#scope:`
  - `WATableCellTag>>#align:`
  - `WATableCellTag>>#character:`
- `WAHtml5ElementsTest` sends `WAHtmlCanvas>>#menu`
- `WALotsaLinksFunctionalTest` sends `WAAnchorTag>>name:`
- `WAPerformanceFunctionalTest` sends `WAHtmlCanvas>>#big` and `WAAnchorTag>>#name:`
- `WAPhraseElementsFunctionalTest` sends `WAHtmlCanvas>>#acronym`
- `WASelectionDateTableFunctionalTest` sends `WATableCellTag>>#align:`

jbrichau:
Removes the use of tags and attributes currently deprecated (or non-functioning) in HTML5 from the Seaside tests and examples.

Also fix the wrong deprecation of:
- the th.scope attribute (it was only deprecated for td tags).
- the charset attribute on meta tags

Fixes #1418

### DIFF
--- a/repository/Seaside-Tests-Functional.package/WAPerformanceFunctionalTest.class/instance/renderAnchorsOn..st
+++ b/repository/Seaside-Tests-Functional.package/WAPerformanceFunctionalTest.class/instance/renderAnchorsOn..st
@@ -4,7 +4,7 @@ renderAnchorsOn: canvas
 		html unorderedList: [
 			html listItem: [
 					html anchor
-						name: 'name';
+						id: 'name';
 						callback: [ ];
 						with: 'Anchor' ] ] ]
 		factor: 1
--- a/repository/Seaside-Canvas.package/WATableHeadingTag.class/instance/scope..st
+++ b/repository/Seaside-Canvas.package/WATableHeadingTag.class/instance/scope..st
@@ -1,0 +1,10 @@
+attributes
+scope: aString
+	"This attribute specifies the set of data cells for which the current header cell provides header information. This attribute may be used in place of the headers attribute, particularly for simple tables. When specified, this attribute must have one of the following values:
+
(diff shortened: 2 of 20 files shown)


repository/Seaside-Tests-Functional.package/WAPerformanceFunctionalTest.class/instance/renderAnchorsOn..st
```diff
@@ -4,7 +4,7 @@ renderAnchorsOn: canvas
 		html unorderedList: [
 			html listItem: [
 					html anchor
-						name: 'name';
+						id: 'name';
```
marschall, line 7: AFAIK this is rendered multiple times. Strictly speaking this leads to invalid HTML as we have multiple element with the same id. I'm fine with either of these:

- Leave as is, it's not a big deal.
- Make it unique.
- Switch to a different attribute like `title`

marschall (commented):
LGTM

repository/Seaside-Tests-Functional.package/WAPerformanceFunctionalTest.class/instance/renderAnchorsOn..st
jbrichau: You're right. Though not really important here, I switched to title.

PR #1444, opened 2024-10-13
Improvements for Web Components

marschall:
Improvements that allow applications with Web Components to run with a lower overhead.

- Save only a single action continuation. Since the back button does not work with Web Components in general we only need to save a single action continuation. This should keep sessions smaller.
- Dedicated action continuation. Since the back button does not work with Web Components in general we can make use of an optimized action continuation that:
  - does not redirect.
  - does not capture state.
  - does not restore state.

In theory these classes could be used in general for "back button less" Seaside applications. However this is quite an advanced topic so I think it's ok to "burry" these in the WebComponents package. Let me know what you think.

### DIFF
--- a/repository/Seaside-WebComponents-Core.package/WASingleElementCache.class/instance/at.ifAbsent..st
+++ b/repository/Seaside-WebComponents-Core.package/WASingleElementCache.class/instance/at.ifAbsent..st
@@ -1,0 +1,5 @@
+accessing
+at: aKey ifAbsent: aBlock
+	^ key = aKey
+		ifTrue: [ value ]
+		ifFalse: [ aBlock value ]
--- a/repository/Seaside-WebComponents-Core.package/WASingleElementCache.class/instance/keyAtValue.ifAbsent..st
+++ b/repository/Seaside-WebComponents-Core.package/WASingleElementCache.class/instance/keyAtValue.ifAbsent..st
@@ -1,0 +1,5 @@
+accessing
+keyAtValue: anObject ifAbsent: aBlock
+	^ value = anObject
+		ifTrue: [ key ]
+		ifFalse: [ aBlock value ]
--- a/repository/Seaside-WebComponents-Core.package/WAWebComponentActionContinuation.class/instance/continue.st
+++ b/repository/Seaside-WebComponents-Core.package/WAWebComponentActionContinuation.class/instance/continue.st
@@ -1,0 +1,4 @@
+processing
+continue
+	"do not capture state"
+	self createRenderContinuation handle: self requestContext
--- a/repository/Seaside-Session.package/WASessionContinuation.class/instance/handle..st
+++ b/repository/Seaside-Session.package/WASessionContinuation.class/instance/handle..st
@@ -2,5 +2,5 @@ handling
 handle: aRequestContext
 	"Resume processing of a request. To ensure valid application state restore all registered states."
 
-	self states restore.
+	self restoreState.
 	self withUnregisteredHandlerDo: [ super handle: aRequestContext ]
--- a/repository/Seaside-Session.package/WASessionContinuation.class/instance/restoreState.st
+++ b/repository/Seaside-Session.package/WASessionContinuation.class/instance/restoreState.st
@@ -1,0 +1,4 @@
+processing
+restoreState
+
+	self states restore
--- a/repository/Seaside-Tests-WebComponents.package/WASingleElementCacheTest.class/instance/createCache.st
+++ b/repository/Seaside-Tests-WebComponents.package/WASingleElementCacheTest.class/instance/createCache.st
@@ -1,0 +1,3 @@
+private
+createCache
+	^ WASingleElementCache new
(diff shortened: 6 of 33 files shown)


adriaon:
Is there a typo or two in https://github.com/marschall/Seaside/blob/master/repository/Seaside-WebComponents-Core.package/WAHeadlessComponent.class/README.md ?

Also, I wouldn't mind if the classes that are useful for SPA's are not buried, but more generally available. Perhaps in a SPA packackage?

marschall:
> Is there a typo or two in https://github.com/marschall/Seaside/blob/master/repository/Seaside-WebComponents-Core.package/WAHeadlessComponent.class/README.md ?

Probably a rendering issue by GitHub. Check out the raw content

https://raw.githubusercontent.com/marschall/Seaside/refs/heads/master/repository/Seaside-WebComponents-Core.package/WAHeadlessComponent.class/README.md

> Also, I wouldn't mind if the classes that are useful for SPA's are not buried, but more generally available. Perhaps in a SPA package?

What classes would you be interested in? `WAWebComponentActionContinuation`? Do you have a suggestion for a name?

> Perhaps in a SPA package?

Maybe, but only if we have more than one class to go there.

adriaon:
> https://raw.githubusercontent.com/marschall/Seaside/refs/heads/master/repository/Seaside-WebComponents-Core.package/WAHeadlessComponent.class/README.md

Got it. Thanks

> What classes would you be interested in? `WAWebComponentActionContinuation`? Do you have a suggestion for a name?
> 
> > Perhaps in a SPA package?
> 
> Maybe, but only if we have more than one class to go there.

If I understand things correctly, i see

- WAWebComponentActionContinuation
- WASingleElementCache
- WASingleContinuationSession

as candidates.

Maybe something like Seaside-SingleContinuation-Core as a name? Or maybe rename Seaside-WebComponents-Core as such?

Cheers,
Adriaan.

repository/Seaside-WebComponents-Core.package/WAWebComponentActionContinuation.class/instance/continue.st
```diff
@@ -0,0 +1,4 @@
+processing
+continue
+	"do not capture state"
```
jbrichau, line 3: Side-note: since we do not capture state, do we need the contents of the `states` instvar in this class?

jbrichau (approved):
Looks good to me. Thanks for this!
I made a couple of code style suggestions in the detail comments.

I was first expecting a "single continuation session" to always reuse the same continuation, like what happens with ajax requests. I do a similar thing in `WATurboRenderPhaseContinuation`. However, I now realise that because the webcomponent re-renders its entire fragment, it's better to throw away the old rendercontext and indeed create an entire new continuation.

repository/Seaside-WebComponents-Core.package/WAWebComponentActionContinuation.class/instance/continue.st
marschall: I see what you meant. The question comes down to how much we want to tie `WAFragmentRenderPhaseContinuation` to `WAFragmentRenderPhaseContinuation`. Right now the `states` inst vars are still filled with `WASnapshot` instances due to two reasons:

- `WAFragmentRenderPhaseContinuation` inherits`WARenderLoopContinuation >> #createActionContinuationWithContext:` and send `#snapshot:renderContext:` to the action phase continuation class
- `WAWebComponentActionContinuation >> #handle:` still sends ends up in `WASessionContinuation >> #handle:`

marschall (commented):
Added responses to review comments

repository/Seaside-WebComponents-Core.package/WASingleElementCache.class/instance/at.ifAbsent..st
```diff
@@ -0,0 +1,5 @@
+accessing
+at: aKey ifAbsent: aBlock
+	^ key = aKey
+		ifTrue: [ value ]
+		ifFalse: [ aBlock value ]
```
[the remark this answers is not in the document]
jbrichau: Yes, I was disregarding that rule as I assumed it was just not smart enough to see that the argument was a block and that avoiding the send of `value` would be better than an optimization.

I was shown now that it's better to not do what I suggested above. So I withdraw my comment.

repository/Seaside-WebComponents-Core.package/WASingleElementCache.class/instance/keyAtValue.ifAbsent..st
```diff
@@ -0,0 +1,5 @@
+accessing
+keyAtValue: anObject ifAbsent: aBlock
+	^ value = anObject
+		ifTrue: [ key ]
+		ifFalse: [ aBlock value ]
```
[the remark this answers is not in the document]
jbrichau: idem as the other comment

repository/Seaside-WebComponents-Core.package/WAWebComponentActionContinuation.class/instance/continue.st
jbrichau: It was not a blocking comment but maybe something to look at further down the road.

jbrichau (approved):

PR #1462, opened 2025-06-29
Camp 2025 Squeak fixes round 1

marschall:
- Skip Zinc streaming tests if Zinc is not loaded
- Use Squeak context API


marschall:
I hope fixes two our Squeak CI failures.

jbrichau (approved):
Good suggestion on the streaming uploads tests. 
Thanks!

PR #1465, opened 2025-06-29
Squeak fixes camp 2025 3

marschall:
- Don't load Seaside-Pharo-Swagger-Core in Squeak, Seaside-Squeak-Swagger-Core is enough.
- Remove #pragmaAt:in:, no senders


marschall:
Depends on #1464

jbrichau:
@marschall I think if you merge the master in, it will work since I updated the Grease dependency.

marschall:
> @marschall I think if you merge the master in, it will work since I updated the Grease dependency.

I restarted the build and it's failing with the same error.

Issue #1445, opened 2024-10-19
Default favicon.ico handler


marschall:
I wanted to ask what peoples option is on adding a default favicon.ico handler, probably in `Seaside-Tools-Web`.

The reason for this is that
- Browsers per default always request at `favicon.ico` resulting in a 404 which shows up when using browser developer tools.
- When using other web application frameworks on port 8080 cached versions of these, eg. the Tomcat logo may be displayed.

The implementation would be relatively simple

```st

WARequestHandler << #WAFaviconHandler

handleFiltered: aRequestContext 
	aRequestContext response
		contentType: (WAMimeType main: 'image' sub: 'x-icon');
		binary;
		nextPutAll: WAToolFiles basicNew faviconIco

class >> initialize
	WADispatcher default register: self new at: 'favicon.ico'

```

marschall:
Fixed by #1478 in 3.6.0

PR #1468, opened 2025-06-30
avoid throwing a 404 if there is no favicon

SabineMa:
https://github.com/SeasideSt/Seaside/issues/1445
avoid throwing a 404 if there is no favicon
avoid showing cached versions of favicon

### DIFF
--- a/repository/Seaside-Tools-Web.package/WAFaviconHandler.class/instance/handleFiltered..st
+++ b/repository/Seaside-Tools-Web.package/WAFaviconHandler.class/instance/handleFiltered..st
@@ -1,0 +1,7 @@
+handling
+handleFiltered: aRequestContext
+
+	aRequestContext response
+		contentType: (WAMimeType main: 'image' sub: 'x-icon');
+		binary;
+		nextPutAll: WAToolFiles basicNew faviconIco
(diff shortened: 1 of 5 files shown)


repository/Seaside-Tools-Web.package/WAFaviconHandler.class/instance/handleFiltered..st
```diff
@@ -0,0 +1,7 @@
+handling
+handleFiltered: aRequestContext
+
+	aRequestContext response
+		contentType: (WAMimeType main: 'image' sub: 'x-icon');
```
jbrichau, line 5: While not necessary, it might make sense to add a convenience method for this mimetype along the way.

repository/Seaside-Tools-Web.package/monticello.meta/categories.st
```diff
@@ -1 +1 @@
-self packageOrganizer ensurePackage: #'Seaside-Tools-Web' withTags: #()!
+SystemOrganization addCategory: #'Seaside-Tools-Web'!
```
jbrichau, line 1: This reverts a change from an earlier format of Pharo. I think we should revert this particular change in this PR.

jbrichau (changes requested):
Good addition, thank you!
I just want the filetree exporter artefact reverted and have a small suggestion that is not obligatory.

SabineMa:
I created a new one because this here was wrong (superclass).

PR #1476, opened 2025-07-03
Microdown renderer

jbrichau:
Adds an ability to render microDown on an html canvas.
This provides the ability to render larger pieces of text easily with microdown syntax.

Only works on Pharo


repository/Seaside-Microdown.package/WAHtmlCanvas.extension/instance/microDown..st
marschall, line 1: Do we also want to provide a method to render a parsed tree for people who store the parse tree rather than "source".

marschall (approved):
I just had a look at the code, I did not check it out. LGTM

It sounds like a cool idea.

What is the intended use case? People have "source" in Microdown and want to render it? Do we also want to supply a method for rendering a parse tree?

repository/Seaside-Microdown.package/WAHtmlCanvas.extension/instance/microDown..st
jbrichau: Did not look at that. The use case I added this for was the development of a new Seaside website where I just wanted to have a lot of text with some layout on a page and did not want to manually implement that. Using microdown was a lot easier.

PR #1481, opened 2025-07-05
JSON api example

jbrichau:
There was no example of a JSON api; to I added one


marschall:
I thought there was something in Swagger tests or examples but can't find it.

PR #1482, opened 2025-07-05
initial version of callback request logger

jbrichau:
This is ongoing work to add better traceability to callbacks when trying to understand a log of requests in a front-end http server for Seaside.


jbrichau:
This was superceded by #1495, #1496 and #1497

PR #1488, opened 2025-08-02
Extend CORS Filter with all possible response headers + allow only a single value in Allow-Origin header

jbrichau:
Extend the `WACORSFilter` with support for the following CORS response headers:
- Access-Control-Allow-Credentials
- Access-Control-Max-Age
- Access-Control-Expose-Headers
- Access-Control-Allow-Headers

While adding tests, I also noticed the implementation for the Allow-Origin header is wrong. It should only ever allow a single value (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin). The implementation itself also only ever produced a single value (the first one in the collection), so I changed the code and adapted the method name; e.g.: #addAllowedOrigin: -> #allowOrigin:

### DIFF
--- a/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinerest..st
+++ b/repository/BaselineOfSeaside3.package/BaselineOfSeaside3.class/instance/baselinerest..st
@@ -48,7 +48,7 @@ baselinerest: spec
 	].
 
 	spec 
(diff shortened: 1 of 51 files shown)


eMaringolo (approved):
Thanks for the changes. They all make sense.

PR #1491, opened 2025-08-19
Fixes sending sort message to the wrong object to display the tally.

lopezca:
Fixes the Space tab (is still slow, but it loads)

image


jbrichau:
Thanks @lopezca !

PR #1493, opened 2025-09-12
Adding GRPackages in order to pass Greese's GRPackageTest

tukanos:
The GRPackageTests>>testPackages relies on GRPackage structure that is not present in the Seaside repository.

I also think that Magritte should test the Seaside integration.  If I will have some tests ready I'll create PR for it too.

I have migrated [Magritte into GemStone](https://github.com/tukanos/magritte).  I have pushed the required GRPackages patches into it.  I have also created a PR in the [upstream Magritte](https://github.com/magritte-metamodel/magritte/pull/388).


tukanos:
@jbrichau  Should I then create a patch for Grease that removes GRPackage completely?

jbrichau:
At the very least, the test is not meant to be executed when other GRPackage dependencies are in the image. So, adding these declarations to Seaside repo is not something we will do.

I remember often wanting to remove them and then reverting it for some reason... but we can try.

PR #1499, opened 2025-10-07
WAWebsocket session accessor update

lopezca:
`WAWebsocket` to inherit from GRObject and provide a session accessor


jbrichau (approved):
Thanks!

Issue #1500, opened 2025-10-08
WAWebSocket touch session fails when the sesion is no longer valid


lopezca:
(updated)
`WAWebSocket` holds a reference to the `WASession` (or subclass) that was used to create it. 

`WAWebSocket>>touchSession` runs every time the websocket receives a message to prevent session expiry, but in some cases, the websocket can persist beyond the session expiry, and If the session has been removed from the cache, then the `parent` slot of the session becomes `nil` and `session application` is therefore `nil` leading to an error.

After some thinking, an improved approach would be to delegate to the session the task of updating its lifetime adequately for the application. This would keep the current implementation in place, but would provide a hook for further customization if desired.

The current implementation would change to

```st
WAWebSocket>> touchSession

    session ifNotNil: [ :s | s touch. ] 

```

and 

```st
WASession>>touch

     self application cache at: self key ifAbsent: [ ]
```

which matches the current implementation.

Issue #1498, opened 2025-10-02
WAWebsocket mixes 2 session variables (slots) which is confusing. Add new getter or overload existing?


lopezca:
`WAWebsocket` inherits from `WAObject`. `WAObject` responds to `session` by trying to get the value from the dynamic variable in the current request context. The websocket does not have this context available when processing messages, but it has a reference to the session variable that was used when creating the socket in the `session` __slot__ of the instance. It does not provide a getter for it and it is only used internally by the `touchSession` message.

If an user wants to have access to the session when dealing with messages from the socket, using `aWAWebSocket session` fails because it is pointing to the wrong place.

I propose to either add a getter `socketSession` to get the value in the slot, or to overload `session` in WAWebSocket to get the value. I can send a PR, but wanted to get feedback on the issue first.

PR #1515, opened 2026-01-21
Delegates session update to the session object.

lopezca:
PR for issue #1500 

- Delegates `touchSession` code to `WASession` and enables specialization if desired.


repository/Javascript-Core.package/JSObject.class/instance/jsStringWith..st
```diff
@@ -2,6 +2,7 @@ utilities
 jsStringWith: aRenderable
 	"Create a renderable that will render the string immediately on the response stream."
 
-	^ (JSString context: self renderContext) 
+	^ (JSString context: self renderContext)
+	      rendererClass: self rendererClass;
```
marschall, line 6: Is this change intentional? It does not look related to the issue at hand.

repository/Seaside-Pharo100-Tools-Spec2.package/WAPharoControlPanel.class/instance/initializeWidgets.st
```diff
@@ -13,19 +13,35 @@ initializeWidgets
...
-						ifFalse: [ 'Stopped' ] ]);
+						ifFalse: [ 'Stopped' ] ])
+					width: 55;
+					yourself);
+		addColumn:
+			(SpStringTableColumn
+				title: 'Serve static files from:'
+				evaluated: [ :adaptor | 
+
+					(adaptor species = ZnZincStaticServerAdaptor)
```
marschall, line 42: Are these changes intentional? It does not look related to the issue at hand.

repository/Seaside-Tests-Core.package/WATestServerAdaptor.class/instance/isStopped.st
```diff
@@ -1,3 +1,4 @@
 testing
 isStopped
-	^ status = #stopped
\ No newline at end of file
+
+	^ status isNil or: [ status = #stopped ]
```
marschall, line 4: When was `status` `nil`?

marschall (commented):
Looks like some unintentional changes made it in.

marschall:
Two questions regarding your approach:

1. Since the web socket is initialized from a request context would it make sense to store the application in an instance variable? Would this simplify the code?
2. Should we close the web socket if the session is expired?

repository/Javascript-Core.package/JSObject.class/instance/jsStringWith..st
marschall: Is that an overlap with #1517?

lopezca:
Apologies, I mixed up the branches and polluted this PR. I'll send a new one.

> Two questions regarding your approach:
> 
>     1. Since the web socket is initialized from a request context would it make sense to store the application in an instance variable? Would this simplify the code?

In my case, using the session to access the application works, but the response to #1498 indicates that there can be sockets that are not attached to a session, and these would benefit from having access to the application object.  

> 
>     2. Should we close the web socket if the session is expired?

That's what I'm currently doing, but I have a session object available. I think there should be a hook in `WAZincWebSocket>>run`

lopezca:
Closed in favor of #1531

repository/Seaside-Tests-Core.package/WATestServerAdaptor.class/instance/isStopped.st
lopezca: I added this adaptor via the UI by mistake and then could not remove it due to status being nil.

PR #1522, opened 2026-03-19
display correct uptime on the status/image tab for Squeak

timrowledge:
Spotted dumb-typo error on the Squeak image status tab; the uptime was being calculated as the time since epoch. Especially daft since the WAImageStatus class already has a proper calculation.
Single method fix in Squeak specific package.


marschall (approved):
Good catch, will have to check Pharo later.

PR #1531, opened 2026-07-05
#1500 delegates session refresh to the session object

lopezca:
Small PR to delegate session refresh from a web socket to the session itself. The changes keep the current implementation logic but allow more complex session refresh code to be modified in subclases of WASession.

### DIFF
--- a/repository/Seaside-Session.package/WASession.class/instance/touch.st
+++ b/repository/Seaside-Session.package/WASession.class/instance/touch.st
@@ -1,0 +1,4 @@
+accessing
+touch 
+
+	self application cache at: self key ifAbsent: [  ]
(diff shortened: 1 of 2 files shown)


lopezca:
@marschall thanks for looking at the other PR, this is the change that I wanted to submit.

repository/Seaside-Session.package/WASession.class/instance/touch.st
```diff
@@ -0,0 +1,4 @@
+accessing
+touch 
+
+	self application cache at: self key ifAbsent: [  ]
```
marschall, line 4: I think it would be preferable to store the application inside the WebSocket like we do with the session and avoid the extension method on `Session`. We already check for a `nil` session, so we could also check for a `nil` applicaiton.

This would also make it easier to close the socket in case the session is expired. It would be great if this could be added. Probably another abstract method.

repository/Seaside-Session.package/WASession.class/instance/touch.st
```diff
@@ -0,0 +1,4 @@
+accessing
+touch 
+
```
marschall, line 3: A comment explaining that in case of an expired session the cache will return `nil` will be helpful.

PR #1537, opened 2026-07-07
Execute scripts rendered by WebComponents

### LINKED ISSUE #1536: Scripts rendered by WebComponents are not executed
`WAWebComponentsLibrary >> #seasideWebComponentsJs` uses [Element.innerHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations) to set the contents of the shadow root to the contents of the response. This does not execute `<script>` elements for security reasons.

> While the property does prevent [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script) elements from executing when they are injected

The problem is known, see
* [Run script tags in innerHTML content](https://www.ghinda.net/article/script-tags/)
* [Executing a <script> tag on append after the DOM has loaded](https://stackoverflow.com/questions/49761241/executing-a-script-tag-on-append-after-the-dom-has-loaded)
* [Script inside shadow dom not working](https://stackoverflow.com/questions/51101473/script-inside-shadow-dom-not-working)

Two main approaches seem to exist:
1. Recreate the scripts using `document.createElement` and reattach them using `Node.appendChild`.
1. Use `Range.createContextualFragment` to create a `DocumentFragment`. This one executes `<script>` elements.

The latter seems easier and preserves the structure of the response so I propose to go with this one.

marschall:
Fixes #1536

### DIFF
--- a/repository/Seaside-WebComponents-Core.package/WAWebComponentsLibrary.class/README.md
+++ b/repository/Seaside-WebComponents-Core.package/WAWebComponentsLibrary.class/README.md
@@ -10,7 +10,7 @@ To use this library
 <html>
   <head>
     <!-- -->
-    <script src="seaside-components.js" defer></script>
+    <script src="seaside-web-components.js" defer></script>
   </head>
   <body>
     <!-- -->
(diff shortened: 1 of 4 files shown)


jbrichau (dismissed):
Looks good to me, though I did not test it.

marschall:
> Looks good to me, though I did not test it.

I struggled with DOM access inside a closed shadow DOM so I went for a `console.info`.

PR #1546, opened 2026-07-28
Add structured syntax suffix support to WAMimeType

adriaon:
Add support for structured syntax suffixes (e.g. application/hal+json and application/atom+xml) to WAMimeType.
Introduces:

- #syntax
- #subName

and updates #isBinary to recognize JSON-based media types via their structured syntax suffix.
Examples:

- application/json
- application/hal+json
- application/problem+json
- application/xml
- application/atom+xml

### DIFF
--- a/repository/Seaside-Core.package/WAMimeType.class/instance/subName.st
+++ b/repository/Seaside-Core.package/WAMimeType.class/instance/subName.st
@@ -1,0 +1,10 @@
+accessing
+subName
+	"Answers the component preceding the final '+' in the sub(type).
+	Answers nil when the sub(type) does not contain a '+'."
+
+	| index |
+
+	index := self subSeparatorIndex.
+
+	^index = 0 ifTrue: [nil] ifFalse: [self sub copyFrom: 1 to: index - 1]
--- a/repository/Seaside-Core.package/WAMimeType.class/instance/syntax.st
+++ b/repository/Seaside-Core.package/WAMimeType.class/instance/syntax.st
@@ -1,0 +1,10 @@
+accessing
+syntax
+	"Answers the structured syntax suffix identified by the final '+' in
+	the sub(type) or, when absent, the sub(type) itself."
+
+	| index |
+
+	index := self subSeparatorIndex.
+
+	^index = 0 ifTrue: [self sub] ifFalse: [self sub copyFrom: index + 1 to: self sub size]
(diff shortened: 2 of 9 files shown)


repository/Seaside-Core.package/WAMimeType.class/instance/syntax.st
```diff
@@ -0,0 +1,10 @@
+accessing
+syntax
+	"Answers the structured syntax suffix identified by the final '+' in
```
marschall, line 3: Where does the word "syntax" come from? [RFC-6838](https://datatracker.ietf.org/doc/html/rfc6838#section-4.2.8) uses the word "suffix". But in this case we should probably answer `nil` if it is missing.

repository/Seaside-Core.package/WAMimeType.class/instance/subName.st
```diff
@@ -0,0 +1,10 @@
+accessing
+subName
```
marschall, line 2: Would `#subPrefix` be preferable?

marschall (commented):
Not 100% sure about the selector names and semantics.

Formatting sometimes a bit different than rest of the code.

repository/Seaside-Core.package/WAMimeType.class/instance/syntax.st
adriaon: Thanks for the feedback. Can you be a bit specific about the formatting? Is there a list of rules?

The problem at hand is to determine whether or not the content type is one of the possible json content types. The names in the rfc and the names used for the parts in a more simple content type like 'application/json' are hard to abstract over.

This implementation depends on a selector that returns 'json' for both 'application/json' and 'application/hal+json'. In 'application/json' there is no suffix and in 'application/hal+json' the suffix is '+json' according to the rfc. A different selector named syntax is used to address this. Do you have a suggestion for an alternative name for this selector?
In 'application/json' there is no sub type name, and there for subName will return nil.

Alternatively, we could avoid abstractions and could do for 'application/json'
- sub --> 'json'
- subName/subPrefix --> nil
- syntax/suffix --> nil
and for 'application/hal+json'
- sub --> 'hal+json'
- subName/subPrefix --> 'hal'
- syntax/suffix --> 'json'
and then implement isJson accordingly using or:.

Or avoid all naming and implement isJson as
self sub = 'json' or: [self sub endsWIthSubcollection: '+json']

What do you think?

